PR #14, opened 2023-02-09
Add option to self bootstrap spegel

### LINKED ISSUE #2: Bootstrapping during outage
Spegel does currently not automatically deal with registry outages, if the registry which has an outage is the same one that the Spegel image is pulled from. The mirror configuration is written by Spegel, and for that to happen the image has to be present on the node.

This becomes a catch 22 scenario where scaling a cluster will be impossible because new nodes will not be able to pull images from the registry with the outage.

We need a solution to protect from these types of situations.

phillebaba:
This change enables Spegel to bootstrap itself by pulling the Spegel image from another Spegel instance. This solves issues that would occur if the GHCR registry were to be down.

Fixes #2


phillebaba:
Closing this in favor of #68 as it is a better solution.

PR #33, opened 2023-02-22
Remove containerd info call when handling manifest request

phillebaba:
It turns out the the info function call was not really needed. All of the info it returns is already present by reading the manifest file itself. I do not think this has any significant performance impact but its a good step towards remove reliance on the Containerd socket.


simongottschlag (approved):
LGTM!

PR #42, opened 2023-03-05
Only use bootstrap function for initial peer discovery

phillebaba:
This change is meant to slightly simplify the bootstrap process of the DHT. I am not however certain that this is the correct method.

### DIFF
--- a/internal/routing/router.go
+++ b/internal/routing/router.go
@@ -32,7 +32,7 @@ type P2PRouter struct {
 }
 
 func NewP2PRouter(ctx context.Context, addr string, b Bootstrapper) (Router, error) {
-	log := logr.FromContextOrDiscard(ctx)
+	log := logr.FromContextOrDiscard(ctx).WithName("p2p")
 
 	h, p, err := net.SplitHostPort(addr)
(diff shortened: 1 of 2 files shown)


internal/routing/router.go
```diff
@@ -76,30 +78,10 @@ func NewP2PRouter(ctx context.Context, addr string, b Bootstrapper) (Router, err
...
-			log.Error(err, "could not connect to bootstrap node", "id", addrInfo.ID)
-			time.Sleep(1 * time.Second)
-			continue
-		}
-		log.Info("established connection with bootstrap node", "id", addrInfo.ID)
-		break
+	errCh := kdht.RefreshRoutingTable()
+	err = <-errCh
+	if err != nil {
+		return nil, fmt.Errorf("could not refersh routing table: %w", err)
```
ahockersten, line 84: Nit
```suggestion
		return nil, fmt.Errorf("could not refresh routing table: %w", err)
```

phillebaba:
Hold off on merging this PR until we get a response on this issue libp2p/go-libp2p-kad-dht/issues/827.

PR #43, opened 2023-03-06
Update `helm` documentation to specify namespace

simongottschlag:
If `--namespace` isn't specified, spegel will
be created in the `default` namespace.

This patch just updates the readme to make sure
someone testing spegel creates the namespace for
it.


phillebaba (approved):
LGTM, thank you!

PR #64, opened 2023-03-14
Improve CID creation by using layer digest

phillebaba:


phillebaba:
Do not merge until impact has been verified with performance tests.

PR #66, opened 2023-03-15
Move mirror configuration logic to run as an init container

phillebaba:
This change moves the mirror configuration from running as part of the main application to running as an init container. This has the minor benefit of not having a host path mount on a long running container. The major reason for doing this is to enable a solution to self bootstrap.

This feature will be implemented in a separate PR. The idea is to have Spegel act as an mutating webhook that mutate any Spegel pods init container to set the image registry to localhost. This solves the bootstrap problem when no Spegel instace exists. When Spegel is up and running the init containers Spegel image will be fetched locally.

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,21 +27,28 @@ import (
 	"github.com/xenitab/spegel/internal/state"
 )
 
-type arguments struct {
+type ConfigurationCmd struct {
+	ContainerdRegistryConfigPath string    `arg:"--containerd-registry-config-path" default:"/etc/containerd/certs.d" help:"Directory where mirror configuration is written."`
 	Registries                   []url.URL `arg:"--registries,required" help:"registries that are configured to be mirrored."`
 	MirrorRegistries             []url.URL `arg:"--mirror-registries,required" help:"registries that are configured to act as mirrors."`
-	ImageFilter                  string    `arg:"--image-filter" help:"inclusive image name filter."`
-	RegistryAddr                 string    `arg:"--registry-addr,required" help:"address to server image registry."`
-	RouterAddr                   string    `arg:"--router-addr,required" help:"address to serve router."`
-	MetricsAddr                  string    `arg:"--metrics-addr,required" help:"address to serve metrics."`
-	ContainerdSock               string    `arg:"--containerd-sock" default:"/run/containerd/containerd.sock" help:"Endpoint of containerd service."`
-	ContainerdNamespace          string    `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
-	ContainerdRegistryConfigPath string    `arg:"--containerd-registry-config-path" default:"/etc/containerd/certs.d" help:"Directory where mirror configuration is written."`
-	ContainerdMirrorAdd          bool      `arg:"--containerd-mirror-add" default:"true" help:"Will add containerd mirror configuration if true."`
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -53,51 +60,54 @@ func main() {
 		panic(fmt.Sprintf("who watches the watchmen (%v)?", err))
 	}
 	log := zapr.NewLogger(zapLog)
+	ctx := logr.NewContext(context.Background(), log)
```
simongottschlag, line 63: I would prefer the notify context to be moved here.

main.go
```diff
@@ -53,51 +60,54 @@ func main() {
...
 
-	err = run(log, args)
+	switch {
+	case args.Configuration != nil:
+		err = configurationCommand(ctx, args.Configuration)
+	case args.Registry != nil:
+		err = registryCommand(ctx, args.Registry)
+	}
```
simongottschlag, line 70: I would prefer having these in their own run() function and then returning in each case and then having a default that always returns an error.

main.go
```diff
@@ -53,51 +60,54 @@ func main() {
...
-			return err
-		}
+	cs, err := pkgkubernetes.GetKubernetesClientset(registryCmd.KubeconfigPath)
+	if err != nil {
+		return err
 	}
-	if args.ContainerdMirrorRemove {
-		defer func() {
-			err = errors.Join(err, mirror.RemoveMirrorConfiguration(ctx, fs, args.ContainerdRegistryConfigPath, args.Registries))
-		}()
```
simongottschlag, line 94: I understand that we're not mounting it anymore and that gives us a smaller surface in the long running container, but I do think it's nice of us to keep this and the mount since people may want to try it out for a short while and then remove spegel and that can potentially cause issues for them. I vote for keeping it and the mount.

simongottschlag (changes requested):
I like it and moves us in the right direction. I see removal of the config volume as something that should be put back and the other comments as preferences but not requirements.

main.go
phillebaba: I have created #67 to track this feature, as it should be implemented in a separate PR.

main.go
```diff
@@ -54,50 +61,58 @@ func main() {
 	}
 	log := zapr.NewLogger(zapLog)
 
-	err = run(log, args)
+	ctx := logr.NewContext(context.Background(), log)
+	ctx, cancel := signal.NotifyContext(ctx, syscall.SIGTERM)
+	defer cancel()
```
simongottschlag, line 66: Move this to the run() function since os.Exit(1) doesn't respect defer.

main.go
```diff
@@ -54,50 +61,58 @@ func main() {
...
-func run(log logr.Logger, args *arguments) (err error) {
-	cs, err := pkgkubernetes.GetKubernetesClientset(args.KubeconfigPath)
+func run(ctx context.Context, args *arguments) error {
+	switch {
+	case args.Configuration != nil:
+		return configurationCommand(ctx, args.Configuration)
+	case args.Registry != nil:
+		return registryCommand(ctx, args.Registry)
+	}
+	return nil
```
simongottschlag, line 83: Return an error since this shouldn't happen.

simongottschlag (approved):
Fix the two comments and then I'm satisfied 👍

main.go
phillebaba: It would not make it past args parse but I added a default anyways.

PR #68, opened 2023-03-16
Enable Spegel scaling during GitHub container registry outage

### LINKED ISSUE #2: Bootstrapping during outage

phillebaba:
This change introduces a mutating webhook in Spegel. The purpose of this webhook is to change the registry of the Spegel image in the init container from `ghcr.io` to `localhost:30021`. When this is done the Spegel image will be pulled from locally within the cluster, instead of from GHCR. 

When Spegel is first installed there will be no instance to mutate any Spegel Pod. This is by design as there is no Spegel instance to serve the image on `localhost:30021`. When Spegel is up an running the mutation will kick in, if there is a Spegel instance that can mutate the image there will be a Spegel image that can be served.

Fixes #2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 
+- [#68](https://github.com/XenitAB/spegel/pull/68) Enable Spegel scaling during GitHub container registry outage. 
+
 ### Changed
 
 - [#82](https://github.com/XenitAB/spegel/pull/82) Filter out localhost from advertised IPs.
--- a/Makefile
+++ b/Makefile
@@ -27,14 +27,6 @@ e2e: docker-build
 	# Pull images onto single node which will never run workload.
 	docker exec kind-worker ctr -n k8s.io image pull docker.io/library/nginx:1.23.0
 	docker exec kind-worker ctr -n k8s.io image pull docker.io/library/nginx@sha256:b3a676a9145dc005062d5e79b92d90574fb3bf2396f4913dc1732f9065f55c4b
-
-	# Deploy Spegel
-	kind load docker-image ${IMG}
-	kubectl --kubeconfig $$KIND_KUBECONFIG create namespace spegel
-	helm --kubeconfig $$KIND_KUBECONFIG upgrade --install --namespace="spegel" spegel ./charts/spegel --set "image.pullPolicy=Never" --set "image.tag=${TAG}"
-	kubectl --kubeconfig $$KIND_KUBECONFIG --namespace spegel rollout status daemonset spegel --timeout 60s
-
-	# Pull images onto single node which will never run workload.
 	docker exec kind-worker ctr -n k8s.io image pull docker.io/library/nginx:1.21.0@sha256:2f1cd90e00fe2c991e18272bb35d6a8258eeb27785d121aa4cc1ae4235167cfd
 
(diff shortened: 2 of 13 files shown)


bittrance (commented):
Elegant solution. Or, in the devops vernacular: "What could possibly go wrong?" :smile: 

I'm wondering if there are any non-Helm gitopsey patterns that will "repair" the init container spec?

Anyway, a few lines of readme might be in order since this is a fundamental feature that users typically wants turned on if they can?

phillebaba:
@bittrance I am not really happy with the solution but I cant really see any other option right now. I know that there are some issues with this solution, one of them is the increased complexity. Regarding non Helm, it is not something I have thought about right now as we only officially support installing with Helm.

phillebaba:
Closing because in the end this is not a good idea and goes against the simplicity goals of Spegel.

PrivatePuffin:
> Closing because in the end this is not a good idea and goes against the simplicity goals of Spegel.

I agree, however I think we can get one step into this direction...
Would you mind adding the option to configure the service to run as type "Loadbalancer" (and preferably also allow its IP settings, labels etc) in the Helm chart?

That way we can set the node to use that as a mirror and set `containerdMirrorAdd: true`

---

Would that solve GHCR outages? Not completely.
But, it would allow using Spegel cached images prior to Spegel being started on the node (caching pre-spegel pulls, including... Spegel itself)

phillebaba:
The challenge is that we need the Spegel  image for the mirror configuration to be added. So even if we have a loadbalancer in front of all Spegel instances the node would not pull from there because there would be no mirror configuration. We already have a fallback service for when the local Spegel instance is down which would direct traffic to a random node in the cluster.

Could you explain further how we could pre cache images with this solutuion?

PrivatePuffin:
"we need the Spegel image for the mirror configuration to be added"

Yes, but I can do that on the host just-fine, I dont *need* Spegel to do that, I do need Spegel to be available.

"We already have a fallback service for when the local Spegel instance is down"

I dont use nodePort much or at-all, so please enlighten, would that be available even when Spegel is not running?

phillebaba:
The ability to configure mirror on the host really varies from provider to provider. Some cloud providers allow you to set this while others don't. I am guessing you are using a provider that would allow you to configure this on the host before it is booted?

The node port service is basically a method for host traffic to reach inside of the cluster CNI network. For the most part there are few differences between a LoadBalancer and a NodePort service. The main difference is that cloud providers will create a backing LoadBalancer with an external IP. You could use the NodePort service instead of making it a LoadBalancer service and probably achieve the same result.

PrivatePuffin:
> The ability to configure mirror on the host really varies from provider to provider. Some cloud providers allow you to set this while others don't. I am guessing you are using a provider that would allow you to configure this on the host before it is booted?
> 
> The node port service is basically a method for host traffic to reach inside of the cluster CNI network. For the most part there are few differences between a LoadBalancer and a NodePort service. The main difference is that cloud providers will create a backing LoadBalancer with an external IP. You could use the NodePort service instead of making it a LoadBalancer service and probably achieve the same result.

Thanks for the note, I only ever used nodePort services for single-node clusters.
Never know they were cluster-widely applied. In that case an LB makes no sense indeed!

PR #92, opened 2023-05-03
Change airgap method and validate pulling non cached images does not work

phillebaba:
The current method is not optimal as dropping the default route has negative effects on the Kind cluster. The new method uses iptables to block non RFC1918 CIDR traffic. A new check has been added to make sure that images cant be pulled from the internet.


bittrance (approved):
Assuming we know that the i/f is always called eth0, this seems like an improvement. Having no default route may indeed generate no-route-to-host in places that do not expect it. You could also do `REJECT` so as to (hopefully) avoid long timeouts.

phillebaba:
As far as I know the interface will always be eth0. Good comment about using reject instead.

Issue #36, opened 2023-03-03
CRI-O support


phillebaba:
Currently Spegel only supports Containerd. CRI-O is another popular container runtime. This should also be supported as an alternative backened. It has the benefit of also forcing the code architecture to not be centered around Containerd, enabling other runtimes to also be implemented.

ksankeerth:
Hi @phillebaba,

I’m interested in contributing to this issue. Since the current implementation focuses only on containerd, there could be some complications. First I could analyze and discuss the solution here.  

Thanks in advance!

phillebaba:
Great! So I focused on Containerd initially because that is what I know best. At first I tried implementing this without a Containerd client but that turned out to be impossible because tag information is not stored in the layers but in a separate boltdb. My other thought was to only use the CRI API as that would keep things compatible with CRI-O but there are endpoints that do not exist in the CRI spec which Spegel needs.

I had a look at CRI-O and could not really find a good client library, might be that I have missed something. I think that is the first step to determine if it is possible.

Then I would probably want to refactor some of the code to use an interface instead of the Containerd client. This would enable us to create different implementations for the image fetching. I could do this to make things easier for you. It would probably simplify testing a lot, I don't really know why I did not do it in the first place.

PR #103, opened 2023-05-17
Refactor oci integration

phillebaba:
Part of #36


phillebaba:
Need to test subscribe in containerd for leaking channels.

PR #125, opened 2023-06-27
Add retry mirroring to new peer if current peer fails

### LINKED ISSUE #1: Mutiple reverse proxy retries
Currently a reverse proxy request is only done once. This means that a failed request will fallback to going out to the source registry to complete the request. It would be good to retry x number of times to different nodes which have the requested resource.

phillebaba:
Currently Spegel will only try to mirror a request once. There are several reasons a request may fail to a peer, ideally it should try the next peer. This change implements a retry logic which tries three different peers before returning an error, on the condition that the peer is returned either before the request times out or the resolve timeout is hit.

Fixes #1

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -135,7 +135,7 @@ func registryCommand(ctx context.Context, args *RegistryCmd) (err error) {
 		return state.Track(ctx, ociClient, router)
 	})
 
-	reg, err := registry.NewRegistry(ctx, args.RegistryAddr, ociClient, router)
+	reg, err := registry.NewRegistry(ctx, args.RegistryAddr, ociClient, router, 3)
```
GenPage, line 138: nitpick: Is it possible to allow to configurable retries instead of defaulting to 3?

PR #127, opened 2023-06-28
Add configuration for resolve retry and timeout

phillebaba:
This change allows end users to change how many times a mirror request should be retried and for how long.

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,15 +34,17 @@ type ConfigurationCmd struct {
 }
 
 type RegistryCmd struct {
-	RegistryAddr            string    `arg:"--registry-addr,required" help:"address to server image registry."`
-	RouterAddr              string    `arg:"--router-addr,required" help:"address to serve router."`
-	MetricsAddr             string    `arg:"--metrics-addr,required" help:"address to serve metrics."`
-	Registries              []url.URL `arg:"--registries,required" help:"registries that are configured to be mirrored."`
-	ContainerdSock          string    `arg:"--containerd-sock" default:"/run/containerd/containerd.sock" help:"Endpoint of containerd service."`
-	ContainerdNamespace     string    `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
-	KubeconfigPath          string    `arg:"--kubeconfig-path" help:"Path to the kubeconfig file."`
-	LeaderElectionNamespace string    `arg:"--leader-election-namespace" default:"spegel" help:"Kubernetes namespace to write leader election data."`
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -34,15 +34,17 @@ type ConfigurationCmd struct {
...
-	LeaderElectionNamespace string    `arg:"--leader-election-namespace" default:"spegel" help:"Kubernetes namespace to write leader election data."`
-	LeaderElectionName      string    `arg:"--leader-election-name" default:"spegel-leader-election" help:"Name of leader election."`
+	RegistryAddr            string        `arg:"--registry-addr,required" help:"address to server image registry."`
+	RouterAddr              string        `arg:"--router-addr,required" help:"address to serve router."`
+	MetricsAddr             string        `arg:"--metrics-addr,required" help:"address to serve metrics."`
+	Registries              []url.URL     `arg:"--registries,required" help:"registries that are configured to be mirrored."`
+	ContainerdSock          string        `arg:"--containerd-sock" default:"/run/containerd/containerd.sock" help:"Endpoint of containerd service."`
+	ContainerdNamespace     string        `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
+	MirrorResolveRetries    int           `arg:"--mirror-resolve-retries" default:"3" help:"Max ammount of mirrors to attempt."`
```
bittrance, line 43: This is not quite about retries? If I set this to a larger number than available nodes, resolution will fail? I don't have a good alternative to suggest, but perhaps something like `--sufficient-ready-mirrors` or `--mirror-resolve-threshold`? (I arrived at this because I assumed retries meant the usual thing and wondered how the retries were spread across the timeout period.)

phillebaba: These two variables basically limit how long Spegel will try to resolve a mirror to forward requests to. Resolving mirrors is async where a channel feeds the next mirror discovered with the desired key. If for some reason the request to the mirror may fail the next mirror found will be attempted. Max retries sets the limit for how many mirrors will be attempted before giving up.

The timeout is similar but enforces a max duration that should be spent resolving a mirror. The reason is simply that it is possible that the key does not exist. In that case we should just give up eventually and return an error.

The reason these are important is because Containerd can be configured with its own timeouts, leading to long waiting duration for keys that don't exist. As Containerd will fallback to the original registry it is better to quickly return rather than spending long amounts of times looking for the mirror.

bittrance: I think I understand now. My assumption was that these two parameters were related (they usually are when making calls over networks) but in fact they are not. I suppose the resolution process is instead a race to see if we can find a live mirror before either threshold is reached. Aight, I can't provide a better name for that. :+1:

PR #134, opened 2023-07-04
Switch to using single service with fallback policy for all requests

phillebaba:
This change moves to using a single service for all registry requests making use of the new traffic policy. This change will require Kubernetes 1.26 so it should not be merged until 1.25 is out of support.


phillebaba:
Need to update architecture in README to reflect the new solution.

PR #145, opened 2023-07-25
Use namespace override to modify Spegel installed namespace

noyoshi:
I would like to be able to use spegel as a subchart in my main helm chart, but I don't want to have all the spegel pods running in the main namespace since it will clutter it up. 

This way, we can chose to modify the namespace where all the spegel stuff goes. By default, it will just use the release namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added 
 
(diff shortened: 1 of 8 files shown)


phillebaba:
@noyoshi I understand why you need this, and it seems like other major Helm charts offer this also. Before I can merge I need you to `make helm-docs` to generate the new README, and also add an entry to the Changelog.

noyoshi:
Will do! Thanks for the quick response.

PR #148, opened 2023-07-27
Add configuration to control if mirror should resolve tag

### LINKED ISSUE #147: imagePullPolicy Always gets cached manifest
Hello,

Thanks for this awesome project :)

We started to PoC spegel in our environments, but discovered that some of our devs were relying on the infamous combo
```
imagePullPolicy: Always
image: myimage:latest
```

it seems that in that case, the pod still gets the old image.

I assume that is because the manifest is also cached and not pulled from the registry every time ?

Can/could we optionally change the behavior and always request the manifests to the registry ?

not sure if that is the right way, but, allowing to switch from :
`capabilities = ["pull", "resolve"]`
to 
`capabilities = ["pull"]`
?
https://github.com/XenitAB/spegel/blob/v0.0.9/internal/oci/containerd.go#L278

Thanks !

phillebaba:
Adds option to disable resovle capability in mirror configuration.

Fixes #147


phillebaba:
Closing in favor of #153

Issue #22, opened 2023-02-10
Single prefer local service


phillebaba:
With the old service topology feature in Kubernetes it was possible to create a Service which would "prefer local". What that means is that a NodePort service could be created which when called upon would always route to the node local Pod if possible. If that were to fail it would route to a Pod in the cluster.  This was a great feature that was sadly left out of the new solution as stated here.

https://discuss.kubernetes.io/t/deprecated-topologykeys-alternative-for-node-locality/18184

This feature may be re introduced in the future, and when that happens we should use it instead of having to mirror registry configurations.

phillebaba:
This is now a stable feature in Kubernetes 1.26 so we should be able to move to using PreferLocal.

https://github.com/kubernetes/enhancements/issues/2086

This will however set a requirement to use Kubernetes 1.26 so it would be a good idea to wait one or two releases until removing support for Kubernetes 1.25.

phillebaba:
This was solved in #1017 and will be the default once all supported Kubernetes versions support the new feature.

PR #162, opened 2023-08-03
Remove X-Spegel-External header

phillebaba:
This is the last required header to be removed. Its purpose was to determine if a request is local or a fallback request from another node in the cluster. The same logic can be implemented by comparing the remote IP to the Node IP instead. Additionally this would have to be done in #22 otherwise, as we would switch to a single registry.

This change simplifies the mirror configuration, so that no extra headers are required. Making it simpler for end users to understand the logic and for further advanced features.


phillebaba:
My logic was wrong about which IP was which. The current solution does not work as the IP when traffic is local is not the node IP.

Issue #67, opened 2023-03-16
Clean up containerd mirror configuration


phillebaba:
Spegel makes persistent changes to the hosts filesystem. This is required to setup the mirror configuration. The mirror configuration should be present on the node for as long as Spegel is being used. A good practice however would be to revert the mirror configuration back to its pre Spegel state if Spegel ever was removed. This is especially useful for those who are just evaluating Spegel but chose not to use it. Keeping the mirror configuration will not break the cluster as it would always fallback to the original registry, and the configuration would be removed when old nodes are removed from the cluster.

The old solution was to add an option to remove the configuration files during shutdown. While a good option it wasn't optimal as it is beneficial to keep the configuration during versions updates of Spegel, as it would fallback to other Spegel instances in the cluster. 

An alternative solution is to create a Helm uninstall hook which would run and remove the configuration from all nodes when the Helm chart is removed. This has the benefit of cleaning up stateful changes and returning the node configuration to the state it existed in before Spegel.

phillebaba:
For clarity I am copying @thejoeejoee comments in #767 here.

> To address this issue, I propose introducing a new command, deconfiguration, which would ensure proper cleanup. I can think of three options to run it:
>
> 1. A new native sidecar container with a non-zero terminationGracePeriodSeconds, which waits for a SIGTERM signal before calling deconfiguration.
> 2. A preStop hook for the registry container, triggering deconfiguration when the container stops.
> 3. Converting the configuration container into a sidecar, setting terminationGracePeriodSeconds, and calling deconfiguration when receiving a SIGTERM signal.
>
> Since the native sidecar container solution requires Kubernetes 1.28 (with stable support in 1.29), I would prioritize option 3 if possible, with option 2 as a fallback.
>
> Something to consider is behavior during a rollout update. Should Spegel be unconfigured when a pod stops and restarts on a node? I believe the answer is yes. Since containerd reloads the configuration automatically, a new pod might introduce a different configuration. Cleaning up during the stop phase would help ensure a consistent and predictable setup.

phillebaba:
@thejoeejoee I think the biggest challenge is that it is not simple to determine if a Pod is being shutdown because the parent resource is being deleted or if the pod is being restarted. In a perfect world there would be some sort of signal for the permanent removal.

I could stand behind an opt in feature to remove the mirror configuration when the pod is removed, and then have it added back on restart. There are probably downsides of doing it this way also.

[...]

PR #170, opened 2023-08-16
Backup existing Containerd mirror configuration

phillebaba:
Currently the mirror configuration write is append only. Meaning that registries are only added, never removed. Some Kubernetes deployments may already contain Containerd mirror configuration. On top of that it turns out that removing registries will not remove the configuration from nodes that have already been configured.

To solve both issues this feature will backup existing configuration on first startup, and then remove all other files. If the backup directory already exists it only remove all other existing configuration files before writing the new configuration.

This also enables us to solve #67 in the future when we figure out how to detect an uninstall properly.

### DIFF
--- a/internal/oci/containerd.go
+++ b/internal/oci/containerd.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"os"
 	"path"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


internal/oci/containerd.go
```diff
@@ -25,6 +26,10 @@ import (
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
+const (
+	backupDir = ".backup"
```
phillebaba, line 30: Need to check if this creates an issue with how containerd reads mirror configuration.

internal/oci/containerd.go
```diff
@@ -25,6 +26,10 @@ import (
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
+const (
+	backupDir = "_backup"
```
phillebaba, line 30: This should not conflict with other registry host names.

PR #171, opened 2023-08-17
Add option to disable resolve

phillebaba:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/xenitab/spegel
 go 1.19
 
 require (
+	github.com/BurntSushi/toml v0.3.1
```
bittrance, line 6: I'm a bit confused. We import BurntSushi's TOML impl, but the actual import below is from `github.com/pelletier/go-toml/v2`, which is indirect. Or is this needed for disambiguation?

phillebaba: That was me beeing to fast and forgetting to run tidy.

PR #181, opened 2023-08-23
Fix mirroring images using index files without a media type

### LINKED ISSUE #172: failed to update all images: failed to walk image manifests: unexpected media type for digest
I installed spegel `0.0.9` from Helm. It started on most of the nodes, but on some it failed:

> `{"level":"error","ts":1691505526.569151,"caller":"build/main.go:68","msg":"","error":"failed to update all images: failed to walk image manifests: unexpected media type  for digest: sha256:17e484fedc7ecaa8c83404aa85f0b615743ce723c05d3735436426f12d9e9c77","stacktrace":"main.main\n\t/build/main.go:68\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}`

I don't know how to troubleshoot this further.

phillebaba:
It turns out that we cannot assume that the media type is set in index files. This was an assumption that is made when walking image descriptors. This change fixes the issue by removing any dependency on reading the media type from index files.

Todo:

- [x] Add fallback in registry for missing media type.
- [x] Add unit test with missing media type.
- [x] Add e2e test with missing media type.

Fixes #172

### DIFF
--- a/internal/oci/containerd.go
+++ b/internal/oci/containerd.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


internal/oci/containerd.go
```diff
@@ -227,10 +223,12 @@ func (c *Containerd) GetBlob(ctx context.Context, dgst digest.Digest) ([]byte, s
...
 		return nil, "", err
 	}
-	if ud.MediaType == "" {
-		return nil, "", fmt.Errorf("blob manifest cannot be empty")
+	if ud.MediaType != "" {
+		return b, ud.MediaType, nil
 	}
-	return b, ud.MediaType, nil
+	// Media type is not a required field in any file. We need a fallback method if the field is not set.
+	logr.FromContextOrDiscard(ctx).Info("using Containerd fallback method to determine media type", "digest", dgst.String())
```
phillebaba, line 230: The last challenge is figuring out what our fallback method should look like. We have a couple of options and none of them are great. The issue is that the media type is set in the "parent descriptor" and we don't have a good method of walking backwards.

Currently the options that I have come up with:

* Keep track of all digests and media types seen when iterating images in a hash map.
* Use Containerd content labels to filter and find parent content.
* Walk all content until parent content is found.
* Use labels on content to do a best guess of content type.
* Attempt to parse content into different structs until one works.

PR #195, opened 2023-09-01
fix spegel namespace in daemonset

noyoshi:
Added more uses of the helper `spegel.namespace` for all namespace related injection, otherwise it will break when installing it in a different namespace as a subchart.


phillebaba (approved):
LGTM

phillebaba:
@noyoshi could you just add a new line in the Changelog?

noyoshi:
Done! @phillebaba

Issue #214, opened 2023-10-01
support for servicemonitor labels, scrape interval and scrape timeout


infa-ddeore:
serviceMonitor object needs to have support to add extra labels for prometheus to discover target, current helm chart doesn't support that

infa-ddeore:
fixed in #220

PR #215, opened 2023-10-01
Support for servicemonitor labels, interval and timeout in helm chart

infa-ddeore:
support for servicemonitor labels, interval and timeout

### DIFF
--- a/charts/spegel/Chart.yaml
+++ b/charts/spegel/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: spegel
 description: Stateless cluster local OCI registry mirror.
 type: application
-version: v0.0.13
-appVersion: v0.0.13
+version: v0.0.1
+appVersion: v0.0.1
--- a/charts/spegel/Chart.yaml
+++ b/charts/spegel/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: spegel
 description: Stateless cluster local OCI registry mirror.
 type: application
-version: v0.0.1
-appVersion: v0.0.1
+version: v0.0.13
+appVersion: v0.0.13
--- a/charts/spegel/templates/servicemonitor.yaml
+++ b/charts/spegel/templates/servicemonitor.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     {{- include "spegel.labels" . | nindent 4 }}
     {{- if .Values.serviceMonitor.labels }}
-    {{ toYaml .Values.serviceMonitor.labels | indent 4}}
+    {{ toYaml .Values.serviceMonitor.labels | nindent 4}}
     {{- end }}    
 spec:
   selector:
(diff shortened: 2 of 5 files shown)


infa-ddeore:
fixes #214

phillebaba:
Thanks for the PR. I just want to check what the use case for having service monitor specific labels is? Could you also add an entry to the Changelog.

infa-ddeore:
thanks for the quick reply @phillebaba, updated Changelog.

Prometheus object installed by prometheus-operator has `serviceMonitorSelector` to watch for only the servicemonitors that match the labels, prometheus in below example will read service monitors matching `mykey: myvalue` label

```
$ k -n monitor get prometheus my-prometheus -o yaml | grep -A 2 serviceMonitorSelector
  serviceMonitorSelector:
    matchLabels:
      mykey: myvalue
```

infa-ddeore:
Updated helm’s chart.yaml with the to be released version

infa-ddeore:
updated helm-docs

charts/spegel/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: spegel
 description: Stateless cluster local OCI registry mirror.
 type: application
-version: v0.0.1
-appVersion: v0.0.1
+version: v0.0.13
+appVersion: v0.0.13
```
phillebaba, line 6: Do not set the versions. They are set during build time along with other variables.

charts/spegel/templates/servicemonitor.yaml
```diff
@@ -6,10 +6,15 @@ metadata:
   namespace: {{ include "spegel.namespace" . }}
   labels:
     {{- include "spegel.labels" . | nindent 4 }}
+    {{- if .Values.serviceMonitor.labels -}}
```
phillebaba, line 9: Use with instead of if here.

phillebaba (commented):
I understand the need for setting servicemonitor specific labels. I have added some comments to the review, if you fix those we can merge.

Just FYI the chart and app version are set dynamically during release. This is mostly done for convenience, on top of that the image digest is also injected in before building the Helm chart.

phillebaba:
Additionally could you please rebase.

charts/spegel/templates/servicemonitor.yaml
infa-ddeore: `with` is used with [predefined key](https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with) but in this case key:value of serviceMonitor specific labels are arbitrary
`range` can be used like this to produce same result

```
  labels:
    {{- include "spegel.labels" . | nindent 4 }}
    {{- range $k, $v := .Values.serviceMonitor.labels }}
    {{ $k }}: {{ $v }}
    {{- end }}
```

let me know if you meant something different for using `with` instead of `if` or `range`

phillebaba: In that case I think keeping the if is better.

infa-ddeore:
i messed up with git rebase in this PR

could you approve #220 which i created from different account?

charts/spegel/templates/servicemonitor.yaml
infa-ddeore: made the changes in #220 PR

charts/spegel/Chart.yaml
infa-ddeore: made the changes in #220 PR

phillebaba:
Closing as #220 is merged.

PR #223, opened 2023-10-10
Propagate closing channel before resolve timeout

phillebaba:
This change aims to address timeouts that occur when a key cannot be resolved at all. Currently every request that can't be resolved will be forced to wait the full timeout. A better solution is to exit early if libp2p closes the channel.

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -179,18 +179,20 @@ func (r *Registry) handleMirror(c *gin.Context, key string) {
 		c.AbortWithError(http.StatusInternalServerError, err)
 		return
 	}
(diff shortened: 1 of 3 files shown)


internal/registry/registry.go
```diff
@@ -221,6 +221,11 @@ func (r *Registry) handleMirror(c *gin.Context, key string) {
 			}
 			log.V(5).Info("mirrored request", "path", c.Request.URL.Path, "url", u.String())
 			return
+		case <-resolveCtx.Done():
```
phillebaba, line 224: Is there a point different between channel closed and resolve context closing?

PR #288, opened 2023-12-10
Move packages to pkg to allow importing

phillebaba:
This change will enable other projects to import Spegel directly instead of deploying it as a Daemonset. The goal is to create a stable API that exposes the required packages without limiting the ability to make changes in the future.

Part of enabling embedding Spegel in k3s.
https://github.com/k3s-io/k3s/pull/8977


brandond:
I have a bunch of stuff queued up, but I'm waiting to make sure I don't need anything else before I submit PRs:
* https://github.com/XenitAB/spegel/compare/main...brandond:spegel:main

phillebaba:
My plan was to wait on merging this until I get your feedback. I have been looking through your changes in k3s and wanted to get a head start on creating a framework for the exposed API. I think it would be best to use functional option pattern to expose and configuration required.  In a perfect world I do not want to expose any dependencies towards libp2p. I think that we can come up with a solution that will work well for all.

brandond:
> In a perfect world I do not want to expose any dependencies towards libp2p.

Unless you wrap all the libp2p options that anyone might want to use, I don't see a good way around that.

phillebaba:
Could we split you changes into a couple of PRs? For example the metrics storage, peer identity, peer storage, and delayed containerd client should be pretty easy to merge. Exporting the bootstrap and OCI package should solve another set of challenges. I just think that if we can agree on how to support an existing router we should be there.

brandond:
Oh yeah I'll split it up when I open PRs.

Are you open to moving everything out of internal or do you want to put a public facade on everything?

phillebaba:
If possible having a public facade would be better, but if it does not work we will just have to expose everything. I pinged you on Slack maybe we can take a longer discussion there?

phillebaba:
Closing as this has been done in separate PRs

PR #306, opened 2023-12-29
Realign the structs in attempt to minimise memory usage.

lukaszraczylo:
Results:
```
spegel/internal/routing/bootstrap.go:20:29: struct with 88 pointer bytes could be 64
spegel/internal/routing/mock.go:8:17: struct with 32 pointer bytes could be 8
spegel/pkg/oci/containerd.go:371:15: struct with 24 pointer bytes could be 16
spegel/internal/registry/registry.go:38:15: struct with 64 pointer bytes could be 40
spegel/main.go:38:18: struct with 200 pointer bytes could be 168
spegel/internal/registry/registry_test.go:65:13: struct with 64 pointer bytes could be 48
spegel/pkg/oci/containerd_test.go:24:13: struct with 64 pointer bytes could be 56
spegel/pkg/oci/containerd_test.go:249:13: struct with 64 pointer bytes could be 56
spegel/pkg/oci/containerd_test.go:281:13: struct of size 96 could be 88
spegel/pkg/oci/image_test.go:12:13: struct with 80 pointer bytes could be 72
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -9,3 +9,4 @@ linters:
     - unused
     - misspell
     - testifylint
+    - fieldalignment
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -9,4 +9,7 @@ linters:
     - unused
     - misspell
     - testifylint
-    - fieldalignment
+linters-settings:
+  govet:
+    enable:
+      - fieldalignment
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -38,10 +38,10 @@ var mirrorRequestsTotal = promauto.NewCounterVec(
 type Registry struct {
 	ociClient        oci.Client
 	router           routing.Router
+	localAddr        string
 	resolveRetries   int
 	resolveTimeout   time.Duration
 	resolveLatestTag bool
-	localAddr        string
 }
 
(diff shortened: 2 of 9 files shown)


phillebaba:
I am guessing you are using https://github.com/dkorunic/betteralign to do this? I have yet to see benchmarks showing the impact on aligning struct properties. I guess it does not hurt to test to see if this has an impact.

Could you also enable fieldalignment in `.golang.ci.yaml` to make sure that we check for this in the future.

```yaml
linters-settings:
  govet:
    enable:
      - fieldalignment
```

lukaszraczylo:
@phillebaba Hi, thank you for prompt response. The fields alignment in general is a 'good practice'. I can tell you that in other projects I'm working on, sticking to strict policy on 'everything needs to be aligned' did magic - see for example https://github.com/go-micro/go-micro/pull/2632

.golangci.yaml
```diff
@@ -8,4 +8,5 @@ linters:
     - staticcheck
     - unused
     - misspell
-    - testifylint
\ No newline at end of file
+    - testifylint
+    - fieldalignment
```
phillebaba, line 12:
```suggestion
linters-settings:
  govet:
    enable:
      - fieldalignment
```

phillebaba (commented):
fieldallignment is not its own linter but part of govet

phillebaba (commented):
LGTM! Thank you :rose:

phillebaba (approved):

PR #309, opened 2024-01-04
fix: label selectors on service monitor and metrics service

bc185174:
Resolves https://github.com/XenitAB/spegel/issues/286.

The two services `spegel` and `spegel-registry` use the same labels. As a result, the `ServiceMonitor` can resolve to the non-metric service `spegel-registry`.


charts/spegel/templates/service.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: {{ include "spegel.fullname" . }}
   namespace: {{ include "spegel.namespace" . }}
   labels:
+    spegel.app.io/service: metrics
```
phillebaba, line 7: Lets change this to `app.kubernetes.io/component: metrics` instead.

phillebaba (commented):
Could you add a Changelog item, rebase and make the change I suggested and we should be good to go.

phillebaba (approved):
LGTM, thank you :rose:

PR #365, opened 2024-02-20
Add support for throttling blob write speed

### LINKED ISSUE #336: rate-limiting on bandwidth

i use k3s , I want to limit the bandwidth of the spegel to a range, if there is no upper limit, it may affect the cluster network and lead to the cluster health

_No response_

phillebaba:
This change will allow users to throttle blob write speed in case nodes are negatively impacted by the increase of IO. This feature is optional and disabled by default.

Fixes #336

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/xenitab/spegel/pkg/registry"
 	"github.com/xenitab/spegel/pkg/routing"
 	"github.com/xenitab/spegel/pkg/state"
+	"github.com/xenitab/spegel/pkg/throttle"
 )
 
 type ConfigurationCmd struct {
@@ -47,17 +48,18 @@ type BootstrapConfig struct {
 
 type RegistryCmd struct {
 	BootstrapConfig
-	ContainerdRegistryConfigPath string        `arg:"--containerd-registry-config-path" default:"/etc/containerd/certs.d" help:"Directory where mirror configuration is written."`
-	MetricsAddr                  string        `arg:"--metrics-addr,required" help:"address to serve metrics."`
-	LocalAddr                    string        `arg:"--local-addr,required" help:"Address that the local Spegel instance will be reached at."`
-	ContainerdSock               string        `arg:"--containerd-sock" default:"/run/containerd/containerd.sock" help:"Endpoint of containerd service."`
-	ContainerdNamespace          string        `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
-	RouterAddr                   string        `arg:"--router-addr,required" help:"address to serve router."`
-	RegistryAddr                 string        `arg:"--registry-addr,required" help:"address to server image registry."`
-	Registries                   []url.URL     `arg:"--registries,required" help:"registries that are configured to be mirrored."`
-	MirrorResolveTimeout         time.Duration `arg:"--mirror-resolve-timeout" default:"5s" help:"Max duration spent finding a mirror."`
-	MirrorResolveRetries         int           `arg:"--mirror-resolve-retries" default:"3" help:"Max amount of mirrors to attempt."`
-	ResolveLatestTag             bool          `arg:"--resolve-latest-tag" default:"true" help:"When true latest tags will be resolved to digests."`
+	BlobSpeed                    *throttle.Byterate `arg:"--blob-speed" help:"Maximum write speed per request when serving blob layers. Should be an integer followed by unit Bps, KBps, MBps, GBps, or TBps."`
+	ContainerdRegistryConfigPath string             `arg:"--containerd-registry-config-path" default:"/etc/containerd/certs.d" help:"Directory where mirror configuration is written."`
+	MetricsAddr                  string             `arg:"--metrics-addr,required" help:"address to serve metrics."`
+	LocalAddr                    string             `arg:"--local-addr,required" help:"Address that the local Spegel instance will be reached at."`
+	ContainerdSock               string             `arg:"--containerd-sock" default:"/run/containerd/containerd.sock" help:"Endpoint of containerd service."`
+	ContainerdNamespace          string             `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
+	RouterAddr                   string             `arg:"--router-addr,required" help:"address to serve router."`
+	RegistryAddr                 string             `arg:"--registry-addr,required" help:"address to server image registry."`
+	Registries                   []url.URL          `arg:"--registries,required" help:"registries that are configured to be mirrored."`
+	MirrorResolveTimeout         time.Duration      `arg:"--mirror-resolve-timeout" default:"5s" help:"Max duration spent finding a mirror."`
+	MirrorResolveRetries         int                `arg:"--mirror-resolve-retries" default:"3" help:"Max amount of mirrors to attempt."`
+	ResolveLatestTag             bool               `arg:"--resolve-latest-tag" default:"true" help:"When true latest tags will be resolved to digests."`
 }
 
 type Arguments struct {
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -47,17 +48,18 @@ type BootstrapConfig struct {
...
-	ContainerdNamespace          string        `arg:"--containerd-namespace" default:"k8s.io" help:"Containerd namespace to fetch images from."`
-	RouterAddr                   string        `arg:"--router-addr,required" help:"address to serve router."`
-	RegistryAddr                 string        `arg:"--registry-addr,required" help:"address to server image registry."`
-	Registries                   []url.URL     `arg:"--registries,required" help:"registries that are configured to be mirrored."`
-	MirrorResolveTimeout         time.Duration `arg:"--mirror-resolve-timeout" default:"5s" help:"Max duration spent finding a mirror."`
-	MirrorResolveRetries         int           `arg:"--mirror-resolve-retries" default:"3" help:"Max amount of mirrors to attempt."`
-	ResolveLatestTag             bool          `arg:"--resolve-latest-tag" default:"true" help:"When true latest tags will be resolved to digests."`
+	BlobSpeed                    *throttle.Byterate `arg:"--blob-speed" help:"Maximum write speed when serving blob layers. Should be number followed by unit Bps, KBps, MBps, GBps, or TBps."`
```
simongottschlag, line 51: Is BlobSpeed meant to be per request or global? My understanding of the code is that it's currently implemented to be per request.

If that's the case, then it should be made clear in the usage of it. If it's meant to be global, maybe we should look into where it's implemented.

main.go
```diff
@@ -176,6 +178,9 @@ func registryCommand(ctx context.Context, args *RegistryCmd) (err error) {
 		registry.WithResolveTimeout(args.MirrorResolveTimeout),
 		registry.WithLocalAddress(args.LocalAddr),
 	}
+	if args.BlobSpeed != nil {
```
simongottschlag, line 181: We have this pattern in two places now. Maybe we should have a No Operation throttle as well and use the noop if BlobSpeed is nil. Then we don't have to care about this anywhere else.

pkg/registry/registry.go
```diff
@@ -294,7 +303,11 @@ func (r *Registry) handleBlob(c *gin.Context, dgst digest.Digest) {
 	if c.Request.Method == http.MethodHead {
 		return
 	}
-	err = r.ociClient.CopyLayer(c, dgst, c.Writer)
+	var writer io.Writer = c.Writer
+	if r.blobSpeed != nil {
+		writer = throttle.NewWriter(c.Writer, *r.blobSpeed)
+	}
```
simongottschlag, line 309: If we have a noop throttle, then we wouldn't have to care about BlobSpeed here.

pkg/throttle/byterate.go
```diff
@@ -0,0 +1,48 @@
...
+	case "Bps":
+		unit = Bps
+	case "KBps":
+		unit = KBps
+	case "MBps":
+		unit = MBps
+	case "GBps":
+		unit = GBps
+	case "TBps":
+		unit = TBps
```
simongottschlag, line 42: Maybe we should write something about these? The difference between KBps, kb/s, Kbps and kbps is always hard.

pkg/throttle/writer.go
```diff
@@ -0,0 +1,37 @@
...
+	"golang.org/x/time/rate"
+)
+
+type writer struct {
+	limiter *rate.Limiter
+	writer  io.Writer
+}
+
+func NewWriter(w io.Writer, br Byterate) io.Writer {
+	burstLimit := 1024 * 1024 * 1024
```
simongottschlag, line 17: Kan we make `1024*1024*1024` a const with a descriptive name?

pkg/throttle/writer.go
```diff
@@ -0,0 +1,37 @@
...
+		writer:  w,
+	}
+}
+
+func (w *writer) Write(p []byte) (int, error) {
+	n, err := w.writer.Write(p)
+	if err != nil {
+		return 0, err
+	}
+	r := w.limiter.ReserveN(time.Now(), n)
```
simongottschlag, line 31: Could be worth it separating out `time.Now()` to be a func on the struct. This could make it possible to write tests where we manipulate time.

simongottschlag (commented):
I like it overall, as long as it's meant to be per request.

main.go
phillebaba: A noop throttler is just extra complexity and call stacks so lets not do that. I think it is better to keep it this way unless we make the argument a pointer instead to allow passing nil pointers.

pkg/throttle/byterate.go
phillebaba: The argument is documented both the command arg and in Helm values. The units follow general nomenclature that B = byte and b = bit. Allowing users to specify Mbps makes no sense as we are writing data in byte chunks.

pkg/throttle/writer.go
phillebaba: The `time.Now()` passed to Reserve the the origin time for the delay duration, and cant be static for all write calls. Don't really see how this would make testing easier?

pkg/throttle/writer.go
phillebaba: I agree.

main.go
phillebaba: Blob speed is per request, implementing a global limit would create a lot more complications with syncing between routines. In my head I would assume that speed would be per request and not global, it is also what all other projects do. I will update the help text to make this clear.

pkg/throttle/byterate.go
```diff
@@ -0,0 +1,48 @@
...
+	default:
+		return fmt.Errorf("unknow unit %s", unitStr)
```
phillebaba, line 44: unknown not unknow

bittrance (approved):
Apart from the timing issue in my comments, I think this implementation is sound. However, I remain unconvinced that this feature is particularly useful. If I want to use this feature to avoid a startup stampede, I probably have to set a low value, but since this is a constant configuration, it will be enforced on every request throughout the spegel process lifetime, even when there are lots of nodes available. That is particularly problematic because in the event of a stampede when there are many nodes in the spegel cluster, I probably want requests to complete as quickly as possible so that the image spreads to other nodes quickly.

Would it not be possible to simply have all requests share the `Limiter` instance? That would probably result in unfair IO scheduling in that it would prioritize high-bandwidth clients, but that might not matter so much, since all clients are "local". The Limiter could be injected with a middleware.

phillebaba:
@irLinja thanks for the input! The routing that currently occurs is best described as random, as the first node found with an image layer is selected. In #336 case the image will initially only exist on a single node, resulting in all new nodes requesting the same image layers at the same time. Containerd by default will also pull layers in parallel which multiplies the amount on concurrent requests.

My understanding was that throttling the write rate of individual request would resolve this issue, but maybe the throttling needs to be implemented at a global scale rather than on a per request basis? If the maximum bandwidth is known we should in that case be able to limit Spegel to only consume a fixed percentage of that right?

irLinja:
> maybe the throttling needs to be implemented at a global scale rather than on a per request basis?
Global scale is my answer.
> If the maximum bandwidth is known we should in that case be able to limit Spegel to only consume a fixed percentage of that right?
Correct but I prefer to limit the total bandwidth to a fixed number rather than using a percentage. Users can then decide if they need a limit and what is a reasonable number depending on their environment.

irLinja:
> I now realize that my logic of a per request throttle would never work as multiple concurrent requests would break this. Going to just refactor this PR to work across multiple requests instead.

What issue are you addressing by implementing write speed limits?
Is this something you still want to add?

phillebaba:
The issue that we are resolving is the risk of a node being overrun with layer fetching like described in the issue. This will most likely affect smaller node in edge deployments rather than large cloud based clusters. Either way this is an optional feature that will not be enabled by default, so I see no harm in implementing it if it is needed.

irLinja:
How is this different from containerd behavior?
If I imagine an edge deployment without Spegel installed, would this be an issue?

> I see no harm in implementing it if it is needed.

Me neither. It's your call, but I'm doubting the need.

If there is a concern with the containerd behavior, stressing the node when pulling many images, then the mirror registry is not the best place to address it.

phillebaba:
The difference with Containerd is that image pulls are done with a limitation to concurrency from a single client. While Spegel could have to serve n time the amount of max concurrent downloads if multiple nodes are pulling from the same node. On top of that I doubt that any registry offers image pull speeds similar to what is possible to image pulls between nodes.

This is why implementing throttling would be a good idea. Would you agree?

irLinja:
Makes sense 🙂

PR #371, opened 2024-02-29
Remove state track retry on error

phillebaba:
PR #343 implemented a retry logic for state tracking. This was for the most important for k3s to make sure that state tracking does not just stop. This had the side effect that state tracking would retry immediately on any error. Before this change when an error occurred, it would be logged and tracking would be attempted again after the ticker.

With the current solution if an image with an invalid name exists on the node the tracking will end up in a infinite retry loop without any delay, which may consume more CPU time than wanted.

@brandond any retry looping should be re implemented in k3s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- [#371](https://github.com/XenitAB/spegel/pull/371) Remove state track retry on error.
+
 ### Fixed
 
 ### Security
--- a/main.go
+++ b/main.go
@@ -167,7 +167,10 @@ func registryCommand(ctx context.Context, args *RegistryCmd) (err error) {
 
 	// State tracking
 	g.Go(func() error {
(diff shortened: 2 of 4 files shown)


brandond:
What about https://github.com/k3s-io/spegel/commit/8fbe436b21588193c9233b6a84c13cd8f15bdad2

phillebaba:
The problem is that the all update returns an error.
https://github.com/XenitAB/spegel/blob/19a2cf04fe9440bb549e5e45ec664ff1d419c4fc/pkg/state/state.go#L45

Which results in the error being logged and Track called immediatly again.
https://github.com/XenitAB/spegel/blob/19a2cf04fe9440bb549e5e45ec664ff1d419c4fc/pkg/state/state.go#L26

The ticker is configured to trigger immediately so any error with all function, like an image parsing error, would result in an immediate trigger again.
https://github.com/XenitAB/spegel/blob/19a2cf04fe9440bb549e5e45ec664ff1d419c4fc/pkg/state/state.go#L31-L37

Your suggested alternative would work but I think it would be better to log any errors from the all function and only return errors when the channel fails. This way the k3s implementation could add its for loop in the its call to Track and Spegel would just error on channel closing unexpectedly.

PR #373, opened 2024-02-29
Implement support to disable registry filtering in Containerd

phillebaba:
Some users may want to mirror all images without specifying a registry. This change modifies the filtering regex to allow for an empty filter registries list. Images will still be filtered however to contain a registry in the reference.

This change would resolve a k3s specific use case for now, but should be supported as an option in Spegel also.
https://github.com/k3s-io/k3s/issues/9590


vflaux:
With no registry configured, no mirror is configured in containerd config.
Shouldn't you add a default configuration to mirror all registries to spegel in this case ? There is a special hostname "_default" for this: https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace.

sfackler:
@phillebaba Is there anything blocking this from merging?

phillebaba:
@sfackler not really, I will rebase and just check my work one more time before merging. Just forgot about this while doing other stuff.

sfackler:
Thanks!

PR #383, opened 2024-03-11
Bump libp2p to v0.33.0, replace deprecated Pretty function

dereknola:
Includes https://github.com/libp2p/go-libp2p/pull/2590 to move towards riscV support

Also automatically `go mod tidy` includes:
https://github.com/XenitAB/spegel/pull/366

Signed-off-by: Derek Nola <derek.nola@suse.com>


phillebaba:
Looks good, could you just add an entry to the changelog and we should be ready to merge.

PR #385, opened 2024-03-13
Update io.copy to io.CopyBuffer to improve copy performance

sonyafenge:
Update io.copy to io.CopyBuffer to improve copy performance:
1. update io.copy to io.CopyBuffer
2. add goroutine to io.copy
3. Add blobCopyBuffer to helmchart

Testing Result:
image


pkg/oci/containerd.go
```diff
@@ -304,7 +305,8 @@ func (c *Containerd) GetManifest(ctx context.Context, dgst digest.Digest) ([]byt
 	return b, mt, nil
 }
 
-func (c *Containerd) CopyLayer(ctx context.Context, dgst digest.Digest, dst io.Writer) error {
+func (c *Containerd) CopyLayer(ctx context.Context, dgst digest.Digest, dst io.Writer, bsize int) error {
```
joejulian, line 308: This _is_ a public function. Is there any backward compatibility guarantees expected here @phillebaba?

pkg/oci/containerd.go
```diff
@@ -314,11 +316,34 @@ func (c *Containerd) CopyLayer(ctx context.Context, dgst digest.Digest, dst io.W
 		return err
 	}
 	defer ra.Close()
-	_, err = io.Copy(dst, content.NewReader(ra))
-	if err != nil {
+
+	// Use a channel to signal the completion of the copy
+	done := make(chan error, 1)
+	// Start a goroutine to perform the copy
+	go func() {
```
phillebaba, line 323: I do not understand why a new go routine is needed here, as we are still waiting for the copy to complete in the end. Is there something I am missing?

pkg/oci/containerd.go
phillebaba: I am a bit hesitant to expose this configuration pattern. If we are going to switch to using copy with buffer we should initially set a default value, unless there is a specific reason to expose it.

As getting the layer size is possible we could have a threshold for the buffer size depending on the amount of data being copied.

phillebaba:
@sonyafenge thank you for benchmarking the difference. I think this is an interesting change if this improves the performance. I just need to understand a bit more the impact of the buffer size and if it is possible to determine the buffer size for the end user. It may be a bit too much to allow each user to set their own buffer size.

Containerd has a Copy helper function that has its own implementation of buffered copy.
https://github.com/containerd/containerd/blob/90ec00c80cbdb4f2f031259a2415f519313fbef2/core/content/helpers.go#L147C35-L147C41

In that implementation it seems to use 1 MB as the buffer size.
https://github.com/containerd/containerd/blob/90ec00c80cbdb4f2f031259a2415f519313fbef2/core/content/helpers.go#L36-L41

My initial thoughts is that we either hard code the buffer size to 1 MB or we switch to using Containerd copy function.

sonyafenge:
> @sonyafenge thank you for benchmarking the difference. I think this is an interesting change if this improves the performance. I just need to understand a bit more the impact of the buffer size and if it is possible to determine the buffer size for the end user. It may be a bit too much to allow each user to set their own buffer size.
> 
> Containerd has a Copy helper function that has its own implementation of buffered copy. https://github.com/containerd/containerd/blob/90ec00c80cbdb4f2f031259a2415f519313fbef2/core/content/helpers.go#L147C35-L147C41
> 
> In that implementation it seems to use 1 MB as the buffer size. https://github.com/containerd/containerd/blob/90ec00c80cbdb4f2f031259a2415f519313fbef2/core/content/helpers.go#L36-L41
> 
> My initial thoughts is that we either hard code the buffer size to 1 MB or we switch to using Containerd copy function.

To reduce the impact to minimum for current user, I use 32KB as default buffer size, which is same size as io.copy. if user don't have specific request for copy speed, they can just use default value and don't need change anything. if user like us care about speed, they can adjust copy buffer based on their hardware, network environment and requirements,

for us, we have enough memory and network bandwidth, we will prefer 2M which is also linux default hugepages size, even more in the future. that's why we need this parameter configable.

pkg/oci/containerd.go
sonyafenge: there is default value: 32768 (32KB), which is the default buffer size for io.copy. I use this value as default value to minimize the impact to current user. User don't need change helm chart value if they are ok with default value. 

> As getting the layer size is possible we could have a threshold for the buffer size depending on the amount of data being copied.

Good idea, this can be thought as a feature request in the future.

sonyafenge:
> My initial thoughts is that we either hard code the buffer size to 1 MB or we switch to using Containerd copy function.

Using containerd copy function also can be a new feature in the future.

pkg/oci/containerd.go
sonyafenge: yes, you are correct, reread the code and found there is already go routine in main.go:
`g.Go(func() error {
		if err := metricsSrv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
			return err
		}
		return nil
	})`
Removed go routine.

phillebaba:
@sonyafenge had time to think about things and have some changes that I would like to make.

BlobCopyBuffer should be a string allowing the unit to be specified. Similar to how byte rate is set. It might be good to move this to a new package called `units` and have both live in that package.
https://github.com/XenitAB/spegel/blob/main/pkg/throttle/byterate.go

The CopyLayer function was probably not a good idea from the get go. It puts the burden of the copy logic on the OCI implementation which could create a similar performance limitation in the future. My suggestion is the we deprecate this function and instead add a new similar method that returns a io.ReadCloser. Then all of this logic would be contained in the registry code instead, which is more logical in my head. Lets call this function `GetBlob` as it compliments `GetManifest`.

```go
func (c *Containerd) GetBlob(ctx context.Context, dgst digest.Digest) (io.ReadCloser, error) {
	client, err := c.Client()
	if err != nil {
		return nil, err
	}
	ra, err := client.ContentStore().ReaderAt(ctx, ocispec.Descriptor{Digest: dgst})
	if err != nil {
		return nil, err
	}
	return struct {
		io.Reader
		io.Closer
	}{
		Reader: content.NewReader(ra),
		Closer: ra,
	}, nil
}
```

Then I am considering if we should change the default value of the buffer size to something larger, I suspect that the majority of users will not dig deep into Spegel and the default value should be a good default. As long as the memory impact is not too great. As I do not know what the impact would be lets leave it as it is and I will come back and change it at a later time based on benchmarks.

sonyafenge:
> @sonyafenge had time to think about things and have some changes that I would like to make.
> 
> BlobCopyBuffer should be a string allowing the unit to be specified. Similar to how byte rate is set. It might be good to move this to a new package called `units` and have both live in that package. https://github.com/XenitAB/spegel/blob/main/pkg/throttle/byterate.go
> 
sorry, I am a little confused, I cannot find the unit package, do you mean you are working on a new package?

sonyafenge:
> Then I am considering if we should change the default value of the buffer size to something larger, I suspect that the majority of users will not dig deep into Spegel and the default value should be a good default. As long as the memory impact is not too great. As I do not know what the impact would be lets leave it as it is and I will come back and change it at a later time based on benchmarks.

I am totally ok to use a larger default value. that's great from my cases. just want to double check is this also target to some edge or IOT users, if yes, larger size may have big impact.

phillebaba:
@sonyafenge I think I have found the root cause of the issue. It was not totally logical.

I had trouble accepting that the issue was in the buffered copy in Go. I have built similar http servers in other programming languages without these types of issues. At the same time it would be weird for Go to move the responsibility on the end user. At the same time other registry projects do not do anything special and use `io.Copy` as it is. My feeling was that it has to be something else.

After some googling and reading a lot of documentation and blog posts about copy performance I realized something was off. The `io.Copy` function has a special case for writers that implement `io.ReaderFrom` that ignore the buffered copy. 
https://github.com/golang/go/blob/1d96895ccbea79daa9d6538bf9da6a4dd197336b/src/io/io.go#L407-L416

At the same time I read mentions that the `http.ResponseWriter` implemented this interface. Specially it implements this interface to make use of Zero copy system calls to eliminate copying data from the kernel to userspace and back into the kernel. This is documented in the `ReadFrom` implementation in the http reponse.
https://github.com/golang/go/blob/1d96895ccbea79daa9d6538bf9da6a4dd197336b/src/net/http/server.go#L574-L576

It turns out that the Gin ResponseWriter is not the same as the http ResponseWriter..... Yay! Which means that the Gin ResponseWriter does not implement the required functions to make use of these system calls. Which results in io.Copy usnig the default buffer instead of using zero copying.

We can test this by comparing the following code, which simulates the same check that the copy function does.

```go
package main

import (
	"fmt"
	"io"
	"net/http"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		if rt, ok := w.(io.ReaderFrom); ok {
			fmt.Printf("Is ReaderFrom: %#v\n\n", rt)
		}
		w.Write([]byte("ok"))
	})
	http.ListenAndServe(":8080", nil)
}
```

```go
package main

import (
	"fmt"
	"io"
	"net/http"

	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()
	r.GET("/", func(c *gin.Context) {
		if rt, ok := c.Writer.(io.ReaderFrom); ok {
			fmt.Printf("Is ReaderFrom: %#v\n\n", rt)
		}
		c.String(http.StatusOK, "ok")
	})
	r.Run(":8080")
}
```
 
This fact is also confirmed in the following Gin PR.
https://github.com/gin-gonic/gin/pull/3197

It is frustrating that the issue is probably from a such simple error. I really do not understand why this is still an issue with Gin considering how long it has been known. I honestly think that the best way forward is to completely remove Gin from Spegel and just use the standard lib instead.

phillebaba:
Great there are blog posts about these issues from the time when I was starting uni. This blog post describes the issue pretty well, and has helped me see this issue in a lot more places.
https://avtok.com/2014/11/05/interface-upgrades.html

I think that the metrics middleware that we are using is also doing a similar extension to capture the status, which may also cause issues even if we remove Gin.

I am aware that we have not verified that this is the source of the issue but I have a good feeling that it is. As I understand the situation better it becomes obvious that there is an issue when bytes are copied through a buffer instead of using a zero copy syscall. These types of issues are probably effecting a lot more projects, but they do not become apparent as they are not serving large amounts of data.

PR #388, opened 2024-03-17
feat: add grafana dashboard to helm chart

onedr0p:
Fixes https://github.com/XenitAB/spegel/issues/387


phillebaba:
@onedr0p I need to check what licensing implications come with including a Grafana dashboard in this repo. I am aware of other projects removing their dashboards after the license change in Grafana occurred.

onedr0p:
> I am aware of other projects removing their dashboards after the license change in Grafana occurred.

Interesting, I don't see that being an issue here though...

onedr0p:
@phillebaba were you able to find out anything? I am guessing you are the one that released the chart on Grafana's dashboard site? https://grafana.com/grafana/dashboards/18089-spegel-stateless-cluster-local-oci-registry-mirror/

phillebaba:
No i have not received answers for my inquiries yet. I am neither the person who created the dashboard published there. I will try to reach out to the person and see if we can get an ok for licensing reasons. 

We need to also verify that this method will work when packaging the chart into an OCI artifact.

phillebaba (approved):
Sorry for taking so long time @onedr0p.

Things look good, I just verified that this would also work when packaged as an OCI artifact. Thanks for the work :rose:

mrclrchtr:
Hi, thanks for adding this.

Unfortunately I get the following message in Grafana when I turn it on:
```
Datasource ${DS_PROMETHEUS} was not found
```
image

What am I doing wrong?

I'm using kube-prometheus-stack.

onedr0p:
@mrclrchtr could you open an issue to track this?

mrclrchtr:
Yes, sure: https://github.com/spegel-org/spegel/issues/509

Issue #312, opened 2024-01-09
Improve configuration tooling


bc185174:
### Describe the problem to be solved

Spegel can be configured through the helm chart values which are propagated to the container arguments; however, there is currently no mechanic to allow end users to configure the application through a config map. This will also be useful when other deployment tooling like kustomizations are used.

### Proposed solution to the problem

Switch to using `github.com/peterbourgon/ff/v3` to parse arguments or env vars.

As an example.

```yaml
        containers:
        - args:
          - --mirror-resolve-retries=3
          - --mirror-resolve-timeout=3s
```

Create a config map:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: configuration
  namespace: spegel
data:
  MIRROR_RESOLVE_RETRIES: "3"
  MIRROR_RESOLVE_TIMEOUT: 3s
```

Consume the config map:

```yaml
        containers:
          envFrom:
          - configMapRef:
              name: configuration
```

phillebaba:
I am not a fan of environment variables for configuration as it forces flat configuration. Your argument has its merits, so the value this would bring could outweigh any future issues. Another idea would be to allow a configuration file instead, but that may take this idea to far for you?

The easy solution is to add environment variable annotations to the existing configuration struct properties.
https://github.com/alexflint/go-arg?tab=readme-ov-file#environment-variables

bc185174:
Generally best to support both (the benefit of `github.com/peterbourgon/ff/v3`).

Flags get pretty messy and pretty fast, especially when you have flags like `--registries` that accept multiple inputs. Flags are not convienient for automation, nor for devs wanting to invoke the binary locally for testing. 

They are useful for interactive-shell invocations. For instance take a tool with two different modes, `monitor` and `info`, you could pass in either `--monitor` or `--info` but often avoided for configuration (other than specifying file path to a configuration file).

1. Support for `ConfigMap` to load in application configuration (common practice for k8s applications). This is ideal if/when start looking at support for `Kustomize` since end-users can declaritevly patch application configuration.
3. Still an option to pass in arguments to the go binary.
4. Ability to create local `.env` file that can be loaded in for local dev testing.
5. Unit-tests can pass in test independant env vars without neededing to exec the binary.

PR #393, opened 2024-03-21
Add environment variable configuration support

kh185193:
## Purpose

Extend spegel's configuration to allow configuration by environment variables.

This allows for more dynamic configuration options, e.g. populating envvars from K8s ConfigMaps or `.env` files.
## Approach

- Extend the configuration struct [go-arg](https://github.com/alexflint/go-arg) annotations to specify envvars for each option.
- Adding `env` tags alongside `arg` flags means either can be used (with [args taking presedence](https://github.com/alexflint/go-arg?tab=readme-ov-file#combining-command-line-options-environment-variables-and-default-values)).

## Issue

 #312


phillebaba:
Thanks for the contribution. I think that this is a reasonable change that has no impact on anyone else. The configuration has grown more than I had expected initially so it might be good to switch to a configuration file later. That will have to be a change for the future.

Issue #378, opened 2024-03-01
Local mirror spend more time than remote download for large image


sonyafenge:
### Spegel version

v0.0.17

### Kubernetes distribution

kubeadm

### Kubernetes version

v1.26.7

### CNI

Calico

### Describe the bug

local mirror download will take more time than remote download image
Repro steps:
1. kubeadm create cluster with 2 worker nodes 
2. install spegel using helm chart 
3. apply a pods yaml with 10GB image to worker-node1, using 3m50s to pulling image
```
Every 2.0s: kubectl describe pod ubuntu-random10g | tail -n 15                                         us-pil1-z1-bastion2: Fri Mar  1 04:49:09 2024

    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 60s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 60s
Events:
  Type     Reason          Age               From     Message
  ----     ------          ----              ----     -------
  Normal   AddedInterface  4m15s             multus   Add eth0 [10.245.6.62/32] from k8s-pod-network
  Normal   Pulling         4m15s             kubelet  Pulling image "penggu/ubuntu:random10G"
  Normal   Pulled          24s               kubelet  Successfully pulled image "penggu/ubuntu:random10G" in 3m50.078812682s (3m50.078824562s includ
ing waiting)
```
5. delete pods 
6. change registry image to private to ensure not pull from remote.
7. apply same pods to worker-node2 using 4m35s to pulling image
```
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 60s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 60s
Events:
  Type     Reason          Age                From     Message
  ----     ------          ----               ----     -------
  Normal   AddedInterface  5m48s              multus   Add eth0 [10.245.239.220/32] from k8s-pod-network
  Normal   Pulling         5m48s              kubelet  Pulling image "penggu/ubuntu:random10G"
  Normal   Pulled          72s                kubelet  Successfully pulled image "penggu/ubuntu:random10G" in 4m35.515593997s (4m35.515608107s inclu
ding waiting)
```

Logs on worker-nodes:
```
[...]

PR #395, opened 2024-03-23
Replace Gin with standard lib HTTP handler

phillebaba:
This change addresses the issues raised by @sonyafenge in #378.

The observations were correct and the data collected really helped me grok the issue at hand. The problem however turned out to be caused by the usage of multiple response writers which caused the writer to not be identified as a reader from during runtime. This in turn meant that Go would fall back to using a in memory buffer when copying blobs to the response writer instead of using a zero  copy system call. While #385 did increase the performance it also did not solve the core problem source. I do however appreciate the effort as it helped understand the problem.

We could have fixed this in Gin, but considering that a PR to do this has been open for over a year an attempt from me would probably have the same effect. Instead I have decided to completely remove Gin and instead implement our own response writer wrapper which does implement the reader from interface. This allows us to decrease the amount of dependencies and remove some complexity. As the registry component is pretty simple I see little use for complex routing logic. 

- [x]  Verify metrics has not changed.
- [x] Verify logs has not changed.

Part of #378


pkg/metrics/metrics.go
```diff
@@ -35,6 +35,21 @@ var (
...
+	HttpResponseSizeHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Subsystem: "http",
+		Name:      "response_size_bytes",
+		Help:      "The size of the HTTP responses.",
+	}, []string{"service", "handler", "method", "code"})
+	HttpRequestsInflight = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "http",
+		Name:      "requests_inflight",
+		Help:      "The number of inflight requests being handled at the same time.",
+	}, []string{"service", "handler"})
```
phillebaba, line 52: Metrics should be documented.

pkg/registry/registry.go
```diff
@@ -91,144 +96,157 @@ func NewRegistry(ociClient oci.Client, router routing.Router, opts ...Option) *R
...
+	defer func() {
+		// Ignore metrics for health endpoint
+		if req.URL.Path == "/healthz" {
+			return
+		}
+
+		// Logging
+		latency := time.Now().Sub(start)
+		path := req.URL.Path
+		kvs := []interface{}{"path", path, "status", rw.Status(), "method", req.Method, "latency", latency, "ip", ""}
```
phillebaba, line 118: IP needs to be added to the logger.

phillebaba:
@sonyafenge do you have the time to verify that this change fixes your issues?

pkg/registry/registry.go
```diff
@@ -258,90 +282,79 @@ func (r *Registry) handleMirror(c *gin.Context, key string) {
...
-	if c.Request.Method == http.MethodHead {
+	rw.Header().Set("Content-Length", strconv.FormatInt(size, 10))
+	rw.Header().Set("Docker-Content-Digest", dgst.String())
+	if req.Method == http.MethodHead {
 		return
 	}
-	var w io.Writer = c.Writer
+	var w io.Writer = rw
 	if r.throttler != nil {
-		w = r.throttler.Writer(c.Writer)
```
phillebaba, line 308: Throttler has the same issue of hiding ReadFrom. At the same time the throttling will have to be set on the reader side instead of writer side for this to actually work.

phillebaba:
I have determined that this change will not fix the issues, it will however bring us closer to the problem. The PR will no longer fix the issue. Hopefully I will be able to solve the read method in Containerd. Either way there is no use delaying merging this.

PR #424, opened 2024-04-08
Add option to append mirror configuration instead of overwriting

### LINKED ISSUE #413: Removal of certs in containerd configuration causing 500 error (mirror resolve retries exhausted for key)
### Spegel version

v0.0.19

### Kubernetes distribution

EKS

### Kubernetes version

v1.27.0

### CNI

VPC CNI

### Describe the bug

We are using jfrog artifactory registry as our registry. we are using tls certificate authentication for resolving and accessing our registry endpoints. We were facing 500 error with mirror resolve retries exhausted for key while running a pod, it gave this error 

 ``` Events:                                                                                                                                                                                                                                                                │
│   Type     Reason     Age                   From               Message                                                                                                                                                                                                 │
│   ----     ------     ----                  ----               -------                                                                                                                                                                                                 │
│   Normal   Scheduled  29m                   default-scheduler  Successfully assigned large-pod-1 to ip-10-190-18-201.eu-west-1.compute.internal                                                                                                                    │
│   Normal   Pulling    27m (x4 over 29m)     kubelet            Pulling image "xxx"                                                                                                  │
│   Warning  Failed     27m (x4 over 29m)     kubelet            Failed to pull image "xxx": rpc error: code = Unknown desc = failed to pull and unpack image "xxx": failed to resolve reference "xxx": unexpected status from HEAD request to http://10.190.18.201:30020/v2/xxx 500 Internal Server Error                                                                                                                                                 │
│   Warning  Failed     27m (x4 over 29m)     kubelet            Error: ErrImagePull                                                                                                                                                                                     │
│   Warning  Failed     27m (x6 over 29m)     kubelet            Error: ImagePullBackOff                                                                                                                                                                                 │
│   Normal   BackOff    4m6s (x109 over 29m)  kubelet            Back-off pulling image xxx"                                                                                         │
│                                                                                                                                                                  
```            

On diving deep we saw that our client.cert and client.key files are not loaded in the configuration              
```
configuration {"level":"info","ts":1712209973.4201605,"caller":"oci/containerd.go:511","msg":"backing up Containerd host configuration","path":"/etc/containerd/certs.d/xxx"}                                                            │
│ configuration {"level":"info","ts":1712209973.420308,"caller":"oci/containerd.go:564","msg":"added containerd mirror configuration","registry":"https://xxx","path":"/etc/containerd/certs.d/xxx/hosts.toml" │
│ configuration {"level":"info","ts":1712209973.4203854,"caller":"oci/containerd.go:564","msg":"added containerd mirror configuration","registry":"xxx","path":"/etc/containerd/certs.d/xxx/hosts.toml" │
│ configuration {"level":"info","ts":1712209973.4204602,"caller":"oci/containerd.go:564","msg":"added containerd mirror configuration","registry":"xxx","path":"/etc/containerd/certs.d/xxx/hosts.toml"}               │
│ configuration {"level":"info","ts":1712209973.4205267,"caller":"build/main.go:87","msg":"gracefully shutdown"}                     
```

only host.toml seems to be loaded to spegel. Therefore not being able to access artifactory and pull.   

[first.log](https://github.com/spegel-org/spegel/files/14863069/first.log)

Sharing the spegel log file for reference. 
Can you please ensure that the all the files in cert configuration is copied in the spegel configuration?

### PR DESCRIPTION
**Pull request by phillebaba**
This change adds a new option to append to existing mirror configuration, which is useful when using Spegel together with other mirrors.

The appending will be done to the contents in the backup directory. As files are only backed up the first time Spegel runs it means that the original mirror configuration will be persisted and reconfigured every time Spegel restarts. The feature is disabled by default to keep the original behavior. 

Fixes #413

### CODE REVIEW

File: pkg/oci/containerd.go [hunk 1]
```diff
@@ -567,6 +566,35 @@ func AddMirrorConfiguration(ctx context.Context, fs afero.Fs, configPath string,
...
+		fp := path.Join(configPath, backupDir, registryURL.Host, "hosts.toml")
+		b, err := afero.ReadFile(fs, fp)
+		if err != nil && !errors.Is(err, afero.ErrFileNotFound) {
+			return hostFile{}, err
+		}
+		if err == nil {
+			hf := hostFile{}
+			err := toml.Unmarshal(b, &hf)
+			if err != nil {
+				return hostFile{}, err
```
phillebaba, line 580: Log when files are appended to instead of created to make it clear to the end user.

pkg/oci/containerd.go
```diff
@@ -536,20 +541,34 @@ func AddMirrorConfiguration(ctx context.Context, fs afero.Fs, configPath string,
...
+			}
+			if err == nil {
+				hf = hostFile{}
+				err := toml.Unmarshal(b, &hf)
+				if err != nil {
+					return err
+				}
+				log.Info("appending to existing Containerd mirror configuration", "registry", registryURL.String(), "path", fp)
+			}
+		}
```
bittrance, line 566: I think I would have formulated this as:
```go
hf := hostFile{
	Server:      server,
	HostConfigs: map[string]hostConfig{},
}
backupUsed := false
if appendToBackup {
	var err
	fp := path.Join(configPath, backupDir, registryURL.Host, "hosts.toml")
	hf, backupUsed, err = loadBackupHostFile(fs, fp)
	if err != nil {
		return err
	}
}
```
Using `backupUsed` we can then generate separate log messages for the separate cases rather than generating an extra log message in the append case.

This leads to three improvements (imo):
- it is clear we read backups from the helper name and the parameter can retain the name`appendMirrors` instead of switching to `appnedToBackup`, and
- it is easier to see that we are replacing `hf` in the append case, and
- denser log output

bittrance (approved):
This looks reasonable. The central code is a little hard to follow, see review comment for suggestion.

pkg/oci/containerd.go
phillebaba: I totally agree with you.

I had the intention of refactoring this code at a later time when implementing the CRIO version. After some trouble understanding my own code I chose to refactor it now. Added your suggestion to reduce the amount of logs.

PR #428, opened 2024-04-09
Add support for CRI-O

phillebaba:
Part of #36 

Depends on https://github.com/cri-o/cri-o/issues/7748


sohankunkerkar:
Hey @phillebaba, https://github.com/cri-o/cri-o/issues/7748 has merged now. Let me know if you need any help with pushing this PR to the finish line.

alexp-openai:
Any updates or an ETA for CRI-O support? Would be good to have it!

phillebaba:
I should probably document this but I have paused this work for now. The reason being that Spegel does not work with CRIO. When I started working on this i was using the storage client to get the individual layers. What I did not know was that unlike Containerd, CRIO never stores the compressed layers. Instead when I was getting the layers I was getting the re-compressed locally. This in turn caused a bunch of problems as it would only work with images that have been compressed using the same compression implementation that CRIO was using.

Until this issue can be resolved I cant really see a way forward.

phillebaba:
Closing as this wont be possible due to the way how CRIO deletes unpacked images and there is no way of reproducing them while keeping the digest the same.

liyibo-tech:
> how CRIO deletes unpacked images

I am also researching the possibility of spegel and crio. Do you mean that because GC deletes the decompressed image, spegel cannot perceive that this image layer has been deleted? If these image layers have been deleted, is it not possible to re-trigger the pull from the remote image repository?

phillebaba:
The challenge is that CRIO deletes compressed layers that have been downloaded after they have been unpacked. Registries serve the compressed layers when a client downloads an image. As far as I know there is no way to disable this behavior in CRIO currently.

Tar encoding is not an idempotent process, meaning that  bits may be slightly different after each execution. Some of this can be avoided by for example disabling metadata headers. Different implementations of tar can produce slightly different files, which would result in the digests not matching. This means that it is impossible to Spegel to re-encode layers as we have no idea which tar implementation the layer was initially created with.

liyibo-tech:
> The challenge is that CRIO deletes compressed layers that have been downloaded after they have been unpacked. Registries serve the compressed layers when a client downloads an image. As far as I know there is no way to disable this behavior in CRIO currently.
> 
> Tar encoding is not an idempotent process, meaning that bits may be slightly different after each execution. Some of this can be avoided by for example disabling metadata headers. Different implementations of tar can produce slightly different files, which would result in the digests not matching. This means that it is impossible to Spegel to re-encode layers as we have no idea which tar implementation the layer was initially created with.
> 
> TLDR; For CRIO to work we need access to compressed layers on disk that have been pulled by CRIO.

@phillebaba  I briefly analysed the github.com/containers/storage/layers.go source code, the Diff function, which uses tar-split technology, the data of the local mirror of crio, each layer has a sha256.tar-split.gz file, according to this file, the image data can be read from the local area, and finally the file in tar.gzip format is generated, and the digest value remains unchanged.

liyibo-tech:
@phillebaba  Can tar-split solve the problem of compressing local files into gzip format and the digest value remains unchanged in the image layer file?

phillebaba:
That may be the case, if I remember correctly this data is returned by re-compressing layers as they are requested. The implementation differs depending on the backing storage.

Its been a while since I gave up on this PR, so things may have changed. What we need is a proof of concept where we are able to fetch a layer with a given digest, that has been previously pulled by CRIO, where the returned content digest matches. If we are able to do that then the rest should be pretty simple to implement in Spegel.

cgwalters:
xref https://github.com/cgwalters/cstor-dist/issues/3

PR #429, opened 2024-04-09
Add metrics to measure duration to resolve peers

phillebaba:
This will help adjusting the default resolve timeout which is currently absurdly high.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -19,6 +19,10 @@ var (
 		Name: "spegel_mirror_requests_total",
 		Help: "Total number of mirror requests.",
 	}, []string{"registry", "cache", "source"})
+	ResolveDurHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "spegel_resolve_duration_seconds",
+		Help: "The duration for router to resolve a peer.",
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -19,6 +19,10 @@ var (
 		Name: "spegel_mirror_requests_total",
 		Help: "Total number of mirror requests.",
 	}, []string{"registry", "cache", "source"})
+	ResolveDurHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "spegel_resolve_duration_seconds",
+		Help: "The duration for router to resolve.",
```
bittrance, line 24: Will this text be included for scraping? If so, it will end up far from its context and might benefit from being a bit more specific. "router" might refer to an HTTP endpoint mux and "resolve" is often about DNS lookups. Inserting e.g. "peer", "cluster" or similar in the description should point the reader in the correct direction.

A similar point can be made about the changelog entry.

bittrance (approved):

pkg/metrics/metrics.go
phillebaba: I do agree and have adjusted it a bit.

PR #436, opened 2024-04-12
Replace http util reverese proxy with custom request forwarding

phillebaba:
The current implementation with reverse proxy was overly complex. Partially this is due to the retry logic implemented, but also the details of how things are implemented in the package. We really do not need all of the features of reverse proxy as we are not forwarding generic requests that could in theory contain any type of data. Instead the requests forwarded are specific to the OCI distribution spec.

This is the reason for replacing reverse proxy with out own http client that forwards requests.

An additional aspect is that the std lib reverse proxy does not use splice or send file when copying the response body. This solution should in theory improve performance.


simongottschlag:
@phillebaba maybe time for a benchmark?

pkg/registry/registry.go
```diff
@@ -255,36 +254,43 @@ func (r *Registry) handleMirror(rw mux.ResponseWriter, req *http.Request, key st
...
 				Host:   ipAddr.String(),
+				Path:   req.URL.Path,
+				// TODO: Should this error early if not set?
+				RawQuery: fmt.Sprintf("ns=%s", req.URL.Query().Get("ns")),
 			}
-			proxy := httputil.NewSingleHostReverseProxy(u)
-			proxy.Transport = r.transport
-			proxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
-				log.Error(err, "proxy failed attempting next")
+			forwardReq, err := http.NewRequest(req.Method, u.String(), nil)
```
simongottschlag, line 268: Use NewRequestWithContext and the req.Context()

phillebaba:
Yup I am working on it, but things need to be worked on a bit. Currently doing it manually. Also using pprof to explore the call stack.

pkg/registry/registry.go
```diff
@@ -32,7 +31,7 @@ type Registry struct {
 	throttler        *throttle.Throttler
 	ociClient        oci.Client
 	router           routing.Router
-	transport        http.RoundTripper
+	httpClient       *http.Client
 	localAddr        string
```
simongottschlag, line 35: Should we initialize it with some sane default?

```go
	httpClient := &http.Client{
		Transport: &http.Transport{
			MaxIdleConns:        20,
			MaxIdleConnsPerHost: 20,
		},
	}
```

pkg/registry/registry.go
```diff
@@ -87,6 +86,7 @@ func NewRegistry(ociClient oci.Client, router routing.Router, opts ...Option) *R
 	r := &Registry{
 		ociClient:        ociClient,
 		router:           router,
+		httpClient:       &http.Client{},
```
simongottschlag, line 89: Should we initialize it with some sane default?

```go
	httpClient := &http.Client{
		Transport: &http.Transport{
			MaxIdleConns:        20,
			MaxIdleConnsPerHost: 20,
		},
	}
```

pkg/registry/registry.go
simongottschlag, line 268: Shouldn't the request body be forwarded? Or do we know that it's always empty?

If we know it's always empty, maybe use `http.NoBody` instead of nil and leave a comment about why.

simongottschlag:
This has been quite helpful for me: https://github.com/XenitAB/go-oidc-middleware/blob/main/.github/workflows/benchmark.yaml

pkg/registry/registry.go
phillebaba: Right now this is a pull only registry so there will never be any request body. If data is sent it is probably wrong.

phillebaba: Worth mentioning that the OCI distribution conformance tests are a part of the e2e tests, which gives some confidence in the changes being made.

phillebaba: Good catch, totally logical thing to do.

phillebaba: // An alternative, however, is to simply set [Request.Body] to nil.

In regards to the question of why not http.NoBody.

phillebaba:
Some testing shows that this does not increase the performance. Either way it is a good change which gives us more control on top of reducing complexity.

phillebaba:
Closing as this is no longer relevant.

PR #443, opened 2024-04-17
FAQ answer for instance failure

### LINKED ISSUE #440: Question: What happens if a spegel node is down?

If a spegel node goes down, then the images stored on it will be unreachable.

Does spegel distribute the images on several hosts, so that an outage of one node does not have an impact?

I read the docs, forgive me if I was blind, but I found nothing about redundant storage of images.

A new part in the documentation which explains what happens if a spegel node is down would be great.

If the answer is "there is no redundancy. If one node goes down, then these image need to be fetched from upstream again", then it is also fine (and  should be part of the docs).

bittrance:
An attempt to explain how the pulling client is affected by a spegel instance failing.

Fixes #440.


phillebaba (approved):
Very detailed explanation of the worst case scenarios.

PR #449, opened 2024-04-20
Replace zapr with slog

phillebaba:
Slog solves all the feature requirements that we so no need to keep zapr.


bittrance (approved):
Indeed, I had a look at logging yesterday and I came to a similar conclusion. I also think that for a focused binary such as spegel, the traditional debug/info/warn/error logging hierarchy works better than the 1-10 scale that logr exposes. I would recommend removing logr and depending on slog directly.

phillebaba:
I would like to remove logr in the future but the issue is that I do like the fact that it is possible to keep the logging object in the context. This is obviously not something that everyone thinks is a good idea. I just think that it is better than a global logger.

PR #452, opened 2024-04-23
Fix Containerd subscribe returning on any error

phillebaba:
The code was copied from a function which kept the return during errors. This was wrong as it would cancel any future event mapping. This change refactors the Subscribe function and resolves the issue. It makes sure to close the channels when the go routine returns to make sure that future errors will bubble up.

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -136,50 +136,49 @@ func verifyStatusResponse(resp *runtimeapi.StatusResponse, configPath string) er
 	return fmt.Errorf("Containerd registry config path is %s but needs to contain path %s for mirror configuration to take effect", cfg.Registry.ConfigPath, configPath)
 }
 
-func (c *Containerd) Subscribe(ctx context.Context) (<-chan ImageEvent, <-chan error) {
-	imgCh := make(chan ImageEvent, 1)
-	errCh := make(chan error, 1)
-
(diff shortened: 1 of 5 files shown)


bittrance (approved):
This is certainly an improvement :+1:  

I'm feeling slightly uncomfortable that the entire tracking logic does log-and-continue. At some point this style turns into a violation of the explode-on-error principle. It is possible to make an argument for an error rate metric here, but let's see if the problem occurs before trying to address it.

pkg/oci/containerd.go
```diff
@@ -136,50 +136,49 @@ func verifyStatusResponse(resp *runtimeapi.StatusResponse, configPath string) er
...
 	client, err := c.Client()
 	if err != nil {
-		errCh <- err
-		close(imgCh)
-		close(errCh)
-		return imgCh, errCh
+		return nil, nil, err
 	}
-
 	envelopeCh, cErrCh := client.EventService().Subscribe(ctx, c.eventFilter)
```
bittrance, line 146: `cErrCh` could do with a better name.

phillebaba: I agree but its so small it should not matter for now.

phillebaba:
I have tried exiting on errors before but that was just as bad in the other way. A crash would occur at the smallest problem with an image, which Spegel can't control. I would rather have things work for the most part than crash for the smallest error.

PR #454, opened 2024-04-24
Update COMPATIBILITY.md

andreyolv:


bittrance:
Can you please provide a section below that explains why kind does not work? Also, please note that we use Kind in our end-to-end tests, see https://github.com/spegel-org/spegel/tree/main/test/e2e, so if you are encountering problems, it would be good to understand why.

phillebaba:
Yes if we are going to add Kind we should add a note about the required Containerd patches which we use in the e2e tests.

```yaml
apiVersion: kind.x-k8s.io/v1alpha4
kind: Cluster
containerdConfigPatches:
- |-
  [plugins."io.containerd.grpc.v1.cri".registry]
    config_path = "/etc/containerd/certs.d"
  # Discarding unpacked layers causes them to be removed, which defeats the purpose of a local cache.
  # Additioanlly nodes will report having layers which no long exist.
  # This is by default false in containerd.
  [plugins."io.containerd.grpc.v1.cri".containerd]
    discard_unpacked_layers = false
  # This is just to make sure that images are not shared between namespaces.
  [plugins."io.containerd.metadata.v1.bolt"]
    content_sharing_policy = "isolated"
```

While an isolated content sharing policy is not required, it makes little sense to run Spegel in kind if it is shared.

@andreyolv is there a specific reason why you want to run Spegel in kind?

andreyolv:
I was not using these plugins for containderd. Thanks for the explanation, I'll try later

The reason for the kind is just to test the tool.

PR #460, opened 2024-04-27
Fix http-bootstrap-addr env variable

stenh0use:
PR fixes the env variable for `HTTPBootstrapAddr` / `--http-bootstrap-addr` flag from `HTTP_BOOTSTRAP_KIND` to `HTTP_BOOTSTRAP_ADDR`. This matches the convention for the other env var options in the file.

While this is a breaking change, I would be surprised if anyone is using the HTTP bootstrap feature yet.

Let me know your thoughts.


phillebaba (approved):
This is my mistake, thanks for fixing this.

PR #471, opened 2024-05-12
Fix handler key in request logging

phillebaba:
During the removal of Gin in #395 I accidentally removed the handler key from the request logs. This key turned out to be very useful for understanding request flows. This change fixes the behavior to work as it did before.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -107,42 +107,46 @@ func (r *Registry) Server(addr string) *http.Server {
 
 func (r *Registry) handle(rw mux.ResponseWriter, req *http.Request) {
 	start := time.Now()
-	handler := req.URL.Path
-	if strings.HasPrefix(handler, "/v2") {
-		handler = "/v2/*"
+	handler := ""
+	path := req.URL.Path
+	if strings.HasPrefix(path, "/v2") {
+		path = "/v2/*"
 	}
-
 	defer func() {
 		latency := time.Since(start)
 		statusCode := strconv.FormatInt(int64(rw.Status()), 10)
 
-		metrics.HttpRequestsInflight.WithLabelValues(handler).Add(-1)
-		metrics.HttpRequestDurHistogram.WithLabelValues(handler, req.Method, statusCode).Observe(latency.Seconds())
-		metrics.HttpResponseSizeHistogram.WithLabelValues(handler, req.Method, statusCode).Observe(float64(rw.Size()))
+		metrics.HttpRequestsInflight.WithLabelValues(path).Add(-1)
+		metrics.HttpRequestDurHistogram.WithLabelValues(path, req.Method, statusCode).Observe(latency.Seconds())
+		metrics.HttpResponseSizeHistogram.WithLabelValues(path, req.Method, statusCode).Observe(float64(rw.Size()))
 
 		// Ignore logging requests to healthz to reduce log noise
 		if req.URL.Path == "/healthz" {
 			return
 		}
 
-		// Logging
-		path := req.URL.Path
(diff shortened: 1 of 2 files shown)


pkg/registry/registry.go
```diff
@@ -160,26 +164,26 @@ func (r *Registry) readyHandler(rw mux.ResponseWriter, req *http.Request) {
 	}
 }
 
-func (r *Registry) registryHandler(rw mux.ResponseWriter, req *http.Request) {
+func (r *Registry) registryHandler(rw mux.ResponseWriter, req *http.Request) string {
 	// Quickly return 200 for /v2 to indicate that registry supports v2.
 	if path.Clean(req.URL.Path) == "/v2" {
-		return
+		return ""
```
bittrance, line 170: Isn't this `version` or something? It is a handler, after all.

pkg/registry/registry.go
```diff
@@ -201,27 +205,30 @@ func (r *Registry) registryHandler(rw mux.ResponseWriter, req *http.Request) {
...
 		r.handleBlob(rw, req, dgst)
-	default:
-		// If nothing matches return 404.
-		rw.WriteHeader(http.StatusNotFound)
+		return "blob"
 	}
+
+	// If nothing matches return 404.
+	rw.WriteHeader(http.StatusNotFound)
+	return ""
```
bittrance, line 231: Would `none`, `unknown` or similar provide better usability (read less support questions)?

pkg/registry/registry.go
```diff
@@ -201,27 +205,30 @@ func (r *Registry) registryHandler(rw mux.ResponseWriter, req *http.Request) {
...
+		return "mirror"
 	}
 
 	// Serve registry endpoints.
 	if dgst == "" {
 		dgst, err = r.ociClient.Resolve(req.Context(), ref)
 		if err != nil {
 			rw.WriteError(http.StatusNotFound, err)
-			return
+			return ""
```
bittrance, line 216: A bit of a nitpick, but from the user's perspective this is actually `manifest` or `blob`. If it turns out that `refType` is "cucumber" it would be better if we ended up on line 231 below than complaining that the "ref" (which is unlikely to be a ref) is unknown.

bittrance (approved):
Clear improvement. Some suggestions to further sharpen the value of the `handler` field.

This makes me wonder if this approach could be extended (in a separate PR) to convey the 404/500 distinction that seems to be confusing users? Strictly speaking, the request log should not be error on 2XX or 4XX.

pkg/registry/registry.go
phillebaba: I think we should keep it as empty for now to keep things simple. If it becomes a problem we can change it in the future.

pkg/registry/registry.go
phillebaba: Yes you are probably right about this.

pkg/registry/registry.go
phillebaba: While this is true the handler is not as important as the error message is included in the error log. I will think about this quickly to see if it is worth making the distinction.

phillebaba:
I know that we have debated it before about error logs and status codes. I have only seen other libraries log 404 status codes as errors. Do you have any example of projects that logs 404 as info level? While we could be "right" is does not help if we are the only ones being "right".

bittrance:
I can go check some common projects and see what they do. This PR is anyway free to ignore that aspect if you feel ready to merge.

pkg/registry/registry.go
phillebaba: I will move this code to the manifest handler later which should resolve this.

PR #475, opened 2024-05-13
Move resolving ref to digest to manifest handler

phillebaba:
Ref resolving should only be done for manifests to it does not make sense to do it for all requests to the registry.


bittrance (commented):
❤️

PR #488, opened 2024-05-16
Update existing registry errors and add more detail

### LINKED ISSUE #484: Improve error reporting

When spegel first pulls an image, the responses from spegel seem to always be 500s which is not very helpful.

- 401/403s on the image pull secrets produce a 500 status code.
- Image not found also produces a 500 error and not a 404 status code.

Improve handling of responses and respond with correct status codes instead of generic 500.

phillebaba:
The current error messages have caused some confusion and unnecessary issues. This change attempts to fix existing errors and add more details to others in the hope that it helps users understand the errors. 

Fixes #484


pkg/registry/registry.go
```diff
@@ -237,23 +239,31 @@ func (r *Registry) handleMirror(rw mux.ResponseWriter, req *http.Request, ref re
...
-			rw.WriteError(http.StatusNotFound, fmt.Errorf("request closed for key: %s", key))
+			rw.WriteError(http.StatusNotFound, fmt.Errorf("mirroring for image component %s has been cancelled: %w", key, resolveCtx.Err()))
 			return
 		case ipAddr, ok := <-peerCh:
 			// Channel closed means no more mirrors will be received and max retries has been reached.
 			if !ok {
-				rw.WriteError(http.StatusNotFound, fmt.Errorf("mirror resolve retries exhausted for key: %s", key))
+				err = fmt.Errorf("mirror with image component %s could not be found", key)
+				if mirrorAttempts > 0 {
+					err = errors.Join(err, fmt.Errorf("requests to %d mirrors failed, all attempts haven been exhausted", mirrorAttempts))
```
bittrance: "haven" -> "have"

pkg/registry/registry.go
```diff
@@ -269,13 +279,11 @@ func (r *Registry) handleMirror(rw mux.ResponseWriter, req *http.Request, ref re
...
 			proxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
-				log.Error(err, "proxy failed attempting next")
+				log.Error(err, "mirror request failed attempting next")
 			}
 			proxy.ModifyResponse = func(resp *http.Response) error {
 				if resp.StatusCode != http.StatusOK {
-					err := fmt.Errorf("expected mirror to respond with 200 OK but received: %s", resp.Status)
-					log.Error(err, "mirror failed attempting next")
-					return err
+					return fmt.Errorf("expected mirror to respond with 200 OK but received: %s", resp.Status)
```
bittrance, line 286: I think this message would be somewhat clearer if it said "peer" rather than "mirror".

pkg/registry/registry.go
```diff
@@ -269,13 +279,11 @@ func (r *Registry) handleMirror(rw mux.ResponseWriter, req *http.Request, ref re
 			proxy := httputil.NewSingleHostReverseProxy(u)
 			proxy.Transport = r.transport
 			proxy.ErrorHandler = func(_ http.ResponseWriter, _ *http.Request, err error) {
-				log.Error(err, "proxy failed attempting next")
+				log.Error(err, "mirror request failed attempting next")
```
bittrance, line 282: Nitpick: this message is ambiguous, and should really be "mirror request failed, attempting next". Alternatively, you could include the new `mirrorAttempts` variable: "request to mirror %d failed".

pkg/registry/registry.go
```diff
@@ -237,23 +239,31 @@ func (r *Registry) handleMirror(rw mux.ResponseWriter, req *http.Request, ref re
...
-			rw.WriteError(http.StatusNotFound, fmt.Errorf("request closed for key: %s", key))
+			rw.WriteError(http.StatusNotFound, fmt.Errorf("mirroring for image component %s has been cancelled: %w", key, resolveCtx.Err()))
 			return
 		case ipAddr, ok := <-peerCh:
 			// Channel closed means no more mirrors will be received and max retries has been reached.
 			if !ok {
-				rw.WriteError(http.StatusNotFound, fmt.Errorf("mirror resolve retries exhausted for key: %s", key))
+				err = fmt.Errorf("mirror with image component %s could not be found", key)
+				if mirrorAttempts > 0 {
+					err = errors.Join(err, fmt.Errorf("requests to %d mirrors failed, all attempts haven been exhausted or timeout has been reached", mirrorAttempts))
```
bittrance, line 259: "haven" -> "have"

bittrance (approved):
LGTM. Minor suggestions.

pkg/registry/registry.go
phillebaba: I agree with this, commas are important. I changed it and removed the attempting next because that would be a lie if max attempts was reached. Instead I changed the log to include the attempt count.

pkg/registry/registry.go
phillebaba: I do not agree with this. The word peer is something that has confused people i other issues. While peer is fine to use in code I would rather use the word mirror externally. It is a word that more people understand in "normal" image mirroring setups.

Issue #472, opened 2024-05-12
Adjust default resolve timeout


phillebaba:
### Describe the problem to be solved

Currently the default resolve timeout is way to high causing poor latency when requesting images that do not exist within the cluster. Five seconds is an unreasonable amount of time to wait for a timeout.

### Proposed solution to the problem

We need to measure and find a good compromise default value that works for most users. It is important that it is not set too low as it would result in the resolve timing out before a peer could be found.

PR #491, opened 2024-05-16
Fix so that resolve timeout does not cancel mirroring attempts

phillebaba:
Using the resolve context as a method of cancelling proxying requests turned out to not be the right answer. When evaluating how to change the default resolve timeout i realized my error that a low enough resolve timeout would cancel any mirror retry even if multiple mirrors has been found. If the resolve timeout was 20ms it would mean that any retry would have to occur before that. In reality it is a lot better to limit the resolve time and populate the peer channel buffer. Then to attempt all peers found in the buffer as long as the client does not timeout.

Part of #472


bittrance (approved):
It might be worthwhile to log explicitly that resolve timeout has occurred?

phillebaba:
It will probably occur a lot an I want to keep the logs clean. Unless someone wants it we can let it be.

PR #494, opened 2024-05-19
Add visualization tool to debug image sources

### LINKED ISSUE #24: Debug source of image
Knowing if Spegel is working or not may be a bit difficult as images fall back to the original registry. For debugging purposes it may be nice to after an image is pulled inspect if it was fetched through Spegel or not. One option could be to annotate or label the image to indicate when it is fetched with Spegel.

phillebaba:
This change will finally empower users to debug Spegel image pulling without having to grok the logs. Over the last year a common question that I receive is "how do I know if Spegel is working?". The install process works sometimes too well which confuses people. While the logs are dynamic and do not tell the whole story of the behavior.

This PR adds a new **disabled by default** debug page which which will visualize image pull events that occurred on the node. The event tracking will for now occur in memory, and will only run when the feature is enabled. This tool is not meant to be used to debug after the fact errors. We will start off small and revisit this feature a couple of more times in the future to expand it.

- [x] Track both outgoing and incoming requests
- [x] Filter outgoing and incoming requests
- [x] Filter based on images
- [x] Add handler to metrics server
- [x] Add option to disable visualizer
- [x] Document feature

Fixes #24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#498](https://github.com/spegel-org/spegel/pull/498) Update to Go 1.22.
 - [#499](https://github.com/spegel-org/spegel/pull/499) Add paralleltest linter and set all unit tests to run in parallel.
 - [#501](https://github.com/spegel-org/spegel/pull/501) Rename mock router to memory router and add tests.
+- [#494](https://github.com/spegel-org/spegel/pull/494) Add visualization tool to debug image sources.
 
 ### Deprecated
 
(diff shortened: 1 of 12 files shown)


phillebaba:
While it works we need some better tool to visualize than graphviz.

image

bittrance:
I looked at event logging yesterday and came up with a similar solution, e.g. not logging HEAD requests and using client IP or peer. The main difference is that in my solution I had:
```diff
@@ -181,9 +197,9 @@ func (r *Registry) registryHandler(rw mux.ResponseWriter, req *http.Request) str
        }

        // Request with mirror header are proxied.
-       if req.Header.Get(MirroredHeaderKey) != "true" {
+       if req.Header.Get(MirroredHeaderKey) == "" {
                // Set mirrored header in request to stop infinite loops
-               req.Header.Set(MirroredHeaderKey, "true")
+               req.Header.Set(MirroredHeaderKey, uuid.New().String())
                r.handleMirror(rw, req, ref)
                return "mirror"
        }
``
The UUID can then be used as the ID in logged events, which will allow connecting events from different spegel peers, effectively working as request tracing.

phillebaba:
I know that Containerd already adds a traceID to its request headers so if we wanted to use it we could. If someone wants this it is better to add this in a new PR. I just feel like adding more debugging features to the logs is the antithesis of this feature.

Issue #459, opened 2024-04-27
Many pods take a long time to be in running state and crash a lot in larger clusters


vitobotta:
### Spegel version

v0.0.22

### Kubernetes distribution

k3s

### Kubernetes version

v1.28.8+k3s1

### CNI

Cilium

### Describe the bug

Hi! First of all thanks a lot for this useful project. It's helping work around problems with pulling images from some nodes in Hetzner Cloud when their IPs have previously been banned by registries for some reason.

I am testing Spegel with a couple of clusters between 200 and 400 nodes, and while many pods are running, many others crash continuously and in the logs I see `attempting to acquire leader lease` followed by `gracefully shutdown`. What am I missing?

Another question I had is related to firewall. At the moment I am not opening any ports in the firewall for Spegel and it seems to work anyway, at least I see all nodes pulling images that are often problematic for the problem explained above. Is this OK or am I supposed to open ports? If I need to open ports, are there any security implications or is the communication between nodes encrypted on that port?

Thanks a lot in advance!

vitobotta:
Update: it seems that only 100 pods are running at the same time. Is this a known limitation?

bittrance:
Hm, if there is no message between `attempting to acquire leader lease` and `shutdown gracefully` that implies that the central context was cancelled, probably by some other part of spegel. There are no other interesting messages in the log? If not, my guess would be that the pod is receiving a SIGTERM from Kubernetes.

Just to be sure: are you consistently seeing exactly 100 pods starting successfully? I did some superficial searching but I cannot find a good candidate for a limit in that range, so I'm wondering if this is the result of configuration in your environment? Some operator injecting a `ResourceQuota` in the namespace perhaps, or configuration for watchers in etcd? If so, there might be some interesting events on the failing pods.

Apart from the registry being a `type: NodePort` service, there is nothing special about spegel inter-pod communication, so it should not normally require firewall configuration. If spegel doesn't know of an image, it will return 404, expecting Kubernetes to find the image by some other means.

vitobotta:
[...]

PR #496, opened 2024-05-20
Fix p2p bootstrap to run on failed readiness check

phillebaba:
It turns out that bootstrap peers does not check the connectivity to any address that it is given. This means that an old lease that has not lapsed may give a peer that does not exist. This means that new pods will bootstrap with the old peer, causing an empty routing table. The issue is not resolved until the startup probe fails and the pod is restarted. I have a hunch that this is the source of the error in #459.

The solution to this problem is either to force a bootstrap on a health probe, or check the bootstrap peer before returning it in the bootstrap function.

### DIFF
--- a/pkg/routing/p2p.go
+++ b/pkg/routing/p2p.go
@@ -135,7 +135,7 @@ func (r *P2PRouter) Close() error {
 	return r.host.Close()
 }
 
-func (r *P2PRouter) Ready() (bool, error) {
+func (r *P2PRouter) Ready(ctx context.Context) (bool, error) {
 	addrInfo, err := r.bootstrapper.Get()
 	if err != nil {
 		return false, err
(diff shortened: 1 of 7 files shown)


phillebaba:
Looks like the default ticker for fix low peers is 2 minutes which explains why it is not run before the startup probe.

https://github.com/libp2p/go-libp2p-kad-dht/blob/913d11cb72ddb5b72eac4969687faf771bb15327/dht.go#L471C1-L493C1

pkg/routing/p2p.go
```diff
@@ -144,6 +144,10 @@ func (r *P2PRouter) Ready() (bool, error) {
 		return true, nil
 	}
 	if r.kdht.RoutingTable().Size() == 0 {
+		err := r.kdht.Bootstrap(ctx)
```
bittrance, line 147: Side effect. Might merit a comment?

bittrance (approved):
The joys of distributed programming. 

All nodes could end up here more or less simultaneously. Will that risk being chatty? Or is the bootstrap process only about re-initializing from incoming gossip?

phillebaba:
All this does is is trigger libp2p to refetch the bootstrap from the bootstrapper and refresh the route table. This is not really that chatty as it will only run when the route table is empty.

PR #507, opened 2024-06-09
Change default resolve timeout to 20ms

### LINKED ISSUE #472: Adjust default resolve timeout

phillebaba:
This change will adjust the default resolve timeout from 5s to 20ms. This will improve pull performance when the image is not present in the cluster. I set this to a high value on purpose a year ago when I knew much less about resolving keys. After creating the benchmarks I realized that the default timeout had a large impact on initial performance.

The new default value comes from tests that I have run with the benchmark on an AKS cluster with 50 nodes. These tests showed that the average resolve time was somewhere between 1ms-2ms with no real extreme values. Setting the default value to 20ms will give some buffer.

Fixes #472


phillebaba:
benchmark-1717934990

Benchmarks show improvement on the v1 tests which is expected. While we do not see any impact on the v2 tests which is also expected.

PR #526, opened 2024-06-26
chore(helm): allow relabeling rules for service monitor

eliasbokreta:
Hello, a small contrib for the helm chart 🙂

## Description
This PR allow a user to provide relabeling rules for the service monitor to inject custom labels, rename or drop others.

~Additional change for `appVersion` to match current latest image version, don't have strong opinion here, feel free to refuse this change if the rest looks good to you.~

## How has this been tested ?
Deployed on our clusters successfully.

## Breaking changes ?
No breaking changes.

### DIFF
--- a/charts/spegel/README.md
+++ b/charts/spegel/README.md
@@ -85,6 +85,8 @@ spec:
 | serviceMonitor.enabled | bool | `false` | If true creates a Prometheus Service Monitor. |
 | serviceMonitor.interval | string | `"60s"` | Prometheus scrape interval. |
 | serviceMonitor.labels | object | `{}` | Service monitor specific labels for prometheus to discover servicemonitor. |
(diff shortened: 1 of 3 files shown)


charts/spegel/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: spegel
 description: Stateless cluster local OCI registry mirror.
 type: application
-version: v0.0.1
-appVersion: v0.0.1
```
phillebaba, line 6: You dont need to update the versions, this is done during build time.

phillebaba (changes requested):

phillebaba:
Small thing, after that we should be ready to merge.

phillebaba (approved):
LGTM :rose:

Issue #549, opened 2024-07-29
Containerd hosts config are sorted and not in the specified order


vflaux:
### Spegel version

v0.0.23

### Kubernetes distribution

na

### Kubernetes version

na

### CNI

na

### Describe the bug

The order of hosts in the hosts.toml files in containerd config generated by the spegel configuration command do not preserve the order of previous config nor the order in arguments.
This is due to the fact that maps marshalled with the go-toml library are sorted by key.
The issue is that the order of the "hosts" table is important in containerd as hosts are tried in order of apparition in the file.

All current tests pass because it seems they take this fact into account.
You can try to switch the order of mirrors in input of a test without changing the expected output. The test will still pass while it shouldn't.

phillebaba:
You are right it seems like the value of the key decides the ordering of the hosts, which is why the tests would pass.

phillebaba:
@vflaux I have thought about it and I think easiest to understand solution is to us Go templates. No need to use embeddings I prefer using a static string to keep things simple. Your initial test to show the problem should be enough to verify things are working as they should. Do you want to do this or can I?

vflaux:
I updated https://github.com/spegel-org/spegel/pull/557 and get ride of the embedded file.

PR #550, opened 2024-07-29
fix containerd config generation

vflaux:
This is proposal to fix https://github.com/spegel-org/spegel/issues/549.
This is a bit complicated in my opinion but I couldn't figure out a better way.
Except by writing a toml file without a toml lib which should be simpler and maybe preferable?

Map are output as toml table sorted by the key, which is not what we expect.
This patch dynamically build a struct with a field for each host (and a toml tag).
This allow us to control the order of hosts in the toml file as go-toml output struct as toml tables while preserving the order of the fields in the struct.

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/containerd/containerd"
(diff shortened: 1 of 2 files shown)


phillebaba:
Thank you for reporting the problem and proposing a solution. I dont really know why I never thought about this being a problem. It is slightly weird that the order of maps actually matter. There is something about your solution that I feel makes things more complicated than it has to be. I will dig a bit and see if there is a easier solution, if not we will merge yours for the time being.

The good thing is that we have tests that validate this for future changes.

vflaux:
I opened a new PR with an alternative implementation using go template : https://github.com/spegel-org/spegel/pull/557.
It's less complicated than this one.

> It is slightly weird that the order of maps actually matter.

That's quite unexpected and a bad choice IMO, they should have used an array instead. Containerd also need to parse the file a second time to get the order : https://github.com/containerd/containerd/blob/7f707b5e7970105723257d483394454049eabe47/core/remotes/docker/config/hosts.go#L519.

phillebaba:
Closing as this has been fixed in #601

PR #553, opened 2024-08-05
feat(helm): re-use resources value for initContainer

t3mi:


phillebaba:
Could you split the init container resource into a separate PR?

PR #557, opened 2024-08-06
fix containerd config generation with go template

vflaux:
This is an alternative proposal to fix https://github.com/spegel-org/spegel/issues/549 using go template instead of a complicated struct generation (https://github.com/spegel-org/spegel/pull/550).
The tests are the same as in https://github.com/spegel-org/spegel/pull/550.

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -1,6 +1,7 @@
 package oci
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -11,6 +12,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"text/template"
 
 	"github.com/containerd/containerd"
 	eventtypes "github.com/containerd/containerd/api/events"
@@ -21,6 +23,7 @@ import (
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pelletier/go-toml/v2"
+	tomlu "github.com/pelletier/go-toml/v2/unstable"
 	"github.com/spf13/afero"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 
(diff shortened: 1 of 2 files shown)


pkg/oci/containerd.go
```diff
@@ -475,3 +504,33 @@ func getHostFile(fs afero.Fs, configPath string, appendToBackup bool, registryUR
 	}
 	return hf, false, nil
 }
+
+// getOrderedHosts parses the given byte slice as TOML data and returns a slice of
+// strings containing the ordered hosts found in the TOML data.
+func getOrderedHosts(b []byte) []string {
```
phillebaba, line 510: Do we still need this? Could this be asserted through string comparison in tests?

phillebaba (changes requested):
Most of this looks good. I like the tests and it solves the problem in a simple way. As long as we remove any use of the toml library from the code I will be fine with merging tthis,

pkg/oci/containerd_test.go
```diff
@@ -466,3 +481,48 @@ func stringListToUrlList(t *testing.T, list []string) []url.URL {
...
+[host.'http://127.0.0.5:5000']
+capabilities = ['pull', 'resolve']
+
+[host.'http://127.0.0.1:5000']
+capabilities = ['pull', 'resolve']
+`)},
+			want: []string{"http://127.0.0.4:5000", "http://127.0.0.3:5000", "http://127.0.0.2:5000", "http://127.0.0.5:5000", "http://127.0.0.1:5000"},
+		},
+	}
+	for _, tt := range tests {
```
phillebaba, line 522: Add `t.Parallel()`

pkg/oci/containerd_test.go
```diff
@@ -466,3 +481,48 @@ func stringListToUrlList(t *testing.T, list []string) []url.URL {
 	}
 	return urls
 }
+
+func TestGetOrderedHosts(t *testing.T) {
```
phillebaba, line 485: Add `t.Parallel()`

pkg/oci/containerd.go
vflaux: Sorry, I don't really understand what you mean.
You need to get the hosts in order to be able to rebuild the config file with the same order of hosts as previously (with spegel hosts added).

vflaux:
> As long as we remove any use of the toml library from the code I will be fine with merging tthis,

The toml lib would still be used to unmarshall the config even if I implement a `getOrderedHosts()` without it and don't use it in the go template.

phillebaba:
Closing as this has been fixed in #601

Issue #548, opened 2024-07-27
Support range requests for blobs


phillebaba:
### Describe the problem to be solved

The spec states that a registry should support range requests according to RFC 9110.

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-blobs

### Proposed solution to the problem

Implement range request support when pulling blobs.

PR #576, opened 2024-09-03
feature support range requests for blobs

t1mt:
#548


pkg/oci/containerd.go
```diff
@@ -294,13 +294,11 @@ func (c *Containerd) GetBlob(ctx context.Context, dgst digest.Digest) (io.ReadCl
...
 		return nil, err
 	}
-	return struct {
-		io.Reader
-		io.Closer
-	}{
-		Reader: content.NewReader(ra),
-		Closer: ra,
-	}, nil
+	rs, err := newHTTPReadSeeker(logr.FromContextOrDiscard(ctx), ra)
```
phillebaba, line 297: The `bytes.Reader` implements `Seek`. We could use it instead which would remove the need for our own seeker implementation.

https://pkg.go.dev/bytes#Reader.Seek

pkg/registry/registry.go
```diff
@@ -336,18 +335,15 @@ func (r *Registry) handleBlob(rw mux.ResponseWriter, req *http.Request, ref refe
...
 		return
 	}
 	defer rc.Close()
-	_, err = io.Copy(w, rc)
-	if err != nil {
-		r.log.Error(err, "error occurred when copying blob")
-		return
-	}
+
+	http.ServeContent(w, req, ref.dgst.String(), time.Time{}, rc)
```
phillebaba, line 346: I remember there being a reason I did not use `ServeContent` initially. Let me look up if the reason I found still holds.

phillebaba (changes requested):

pkg/registry/registry.go
phillebaba: Size no longer needs to be set at the beginning of the function as it can now be determined by seek. Also set an empty string as the name.

pkg/oci/httpreadseeker.go
```diff
@@ -0,0 +1,124 @@
+package oci
+
+import (
+	"io"
+
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/errdefs"
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
```
phillebaba, line 9: Remove use of this package.

pkg/oci/httpreadseeker.go
```diff
@@ -0,0 +1,124 @@
...
+	offset             int64
+	errsWithNoProgress int
+	closed             bool
+}
+
+var _ io.ReadSeekCloser = (*httpReadSeeker)(nil)
+
+func newHTTPReadSeeker(log logr.Logger, reader content.ReaderAt) (io.ReadSeekCloser, error) {
+	if reader == nil {
+		return nil, errors.Errorf("httpReadSeeker: reader cannot be nil")
```
phillebaba, line 27:
```suggestion
		return nil, errors.New("reader cannot be nil")
```

pkg/oci/httpreadseeker.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return &httpReadSeeker{
+		log:  log,
+		rc:   reader,
+		size: reader.Size(),
+	}, nil
+}
+
+func (hrs *httpReadSeeker) Read(p []byte) (n int, err error) {
```
phillebaba, line 36: Remove use of named returns.

pkg/oci/httpreadseeker.go
```diff
@@ -0,0 +1,124 @@
...
+		// connection closed unexpectedly. try reconnecting.
+		if n == 0 {
+			hrs.errsWithNoProgress++
+			if hrs.errsWithNoProgress > maxRetry {
+				return // too many retries for this offset with no progress
+			}
+		}
+		if hrs.rc != nil {
+			if clsErr := hrs.rc.Close(); clsErr != nil {
+				hrs.log.Error(clsErr, "httpReadSeeker: failed to close ReadCloser")
```
phillebaba, line 56: We dont need the logger. We should just return the error.

pkg/oci/httpreadseeker.go
```diff
@@ -0,0 +1,124 @@
...
+	if hrs.rc != nil {
+		return hrs.rc.Close()
+	}
+
+	return nil
+}
+
+func (hrs *httpReadSeeker) Seek(offset int64, whence int) (int64, error) {
+	if hrs.closed {
+		return 0, errors.Errorf("Fetcher.Seek: closed: %v", errdefs.ErrUnavailable)
```
phillebaba, line 89: No need to use Containerd errors here.

phillebaba:
One suggestion is to use `SectionReader` to convert the `ReaderAt` to a `Seeker`.

https://pkg.go.dev/io#SectionReader

While the intention is to allow reading specific sections in a file we can use it to pass the full offset. It would remove the need to implement our own seeker.

pkg/oci/oci_test.go
```diff
@@ -171,7 +171,6 @@ func TestOCIClient(t *testing.T) {
 					} else {
 						rc, err := ociClient.GetBlob(ctx, tt.dgst)
 						require.NoError(t, err)
-						defer rc.Close()
```
phillebaba, line 174: Why remove the deferred close here?

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## v0.0.26
 
+### Added
+
+- [#576](https://github.com/spegel-org/spegel/pull/576) Add support for range requests for blobs.
```
phillebaba, line 35: Move this up to unreleased.

internal/mux/response.go
```diff
@@ -15,6 +15,12 @@ type ResponseWriter interface {
 	Size() int64
 }
 
+func NewResponseWriter(w http.ResponseWriter) ResponseWriter { //nolint:ireturn // Return a pointer
```
phillebaba, line 18: This looks like unused code. If it is it should be removed.

pkg/registry/registry.go
```diff
@@ -341,18 +340,14 @@ func (r *Registry) handleBlob(rw mux.ResponseWriter, req *http.Request, ref refe
 	if req.Method == http.MethodHead {
 		return
 	}
-	var w io.Writer = rw
+	var w mux.ResponseWriter = rw
```
phillebaba, line 343: This can be removed and we can pass rw right into http.ServeContent. I forgot to remove this while doing some refactoring.

phillebaba (commented):
Great job, some small fixes and we should be good to merge this :rose:

phillebaba (approved):
LGTM! Thank you :rose:

PR #581, opened 2024-09-10
Skip status response verification for containerd v2

### LINKED ISSUE #579: `verifyStatusResponse` does not work with containerd v2
### Spegel version

v0.0.24

### Kubernetes distribution

Talos

### Kubernetes version

v1.31.0

### CNI

Cilium

### Describe the bug

After updating my homelab cluster to Talos v1.8.0-beta0, I noticed that the spegel pod no longer launched. The logs just say

```
{"time":"2024-09-10T07:29:25.425786168Z","level":"ERROR","source":{"function":"main.main","file":"/build/main.go","line":89},"msg":"run exit with error","err":"Containerd registry config path needs to be set for mirror configuration to take effect"}
```

before the program terminates. Looking at the code, that's `pkg.oci.containerd verifyStatusResponse()` failing its check for `ConfigPath` in the received containerd config.

https://github.com/spegel-org/spegel/blob/main/pkg/oci/containerd.go#L126

I have a feeling Talos is not configuring containerd correctly, but the config file reads:

```toml
version = 3

[plugins]
  [plugins.'io.containerd.cri.v1.images']
    discard_unpacked_layers = false

    [plugins.'io.containerd.cri.v1.images'.registry]
      config_path = '/etc/cri/conf.d/hosts'
```

which according to https://github.com/containerd/containerd/blob/main/docs/cri/config.md is the correct way in 2.x.

jfroy:
containerd v2 refactored its config structures/objects and the configuration that spegel is verifying in `verifyStatusResponse` is no longer returned. This patch adds a version check to `Verify` to allow spegel to skip the verification and take it on blind faith that it is configured correctly to work.

Fixes #579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [#581](https://github.com/spegel-org/spegel/pull/581) Skip status response verification for containerd v2
+
 ### Security
 
 ## v0.0.24
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	semver "github.com/Masterminds/semver/v3"
 	"github.com/containerd/containerd"
 	eventtypes "github.com/containerd/containerd/api/events"
 	"github.com/containerd/containerd/content"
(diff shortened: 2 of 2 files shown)


phillebaba:
@jfroy could you rebase and add an entry to the changelog and we should be ok to merge.

jfroy:
> @jfroy could you rebase and add an entry to the changelog and we should be ok to merge.

Sure thing, will do later tonight (PST).

phillebaba (approved):
LGTM :rose:

PR #612, opened 2024-10-17
Update Spegel Grafana dashboard with Spegel-specific metrics

AhmedTremo:
# Why updated Grafana dashboard?

- The old dashboard included general Kubernetes metrics such as node statuses, running containers, pods, Kubelet operations, which was making it hard to follow Spegel-specific metrics.
- The old dashboard wasn't function properly and didn't include time-series panels to monitor change over time.

By concentrating on Spegel-specific metrics, users can more effectively monitor and troubleshoot Spegel without clutter from other Kubernetes metrics (I think they should have their own dashboard).

This is a view of the new dashboard in action. Let me know your thoughts

1-correct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [#608](https://github.com/spegel-org/spegel/pull/608) Use custom proxy transport and increase idle connections per host.
+- [#612](https://github.com/spegel-org/spegel/pull/612) Update Spegel Grafana dashboard with improved Spegel-specific metrics and layout.
 
 ### Deprecated
 
--- a/charts/spegel/monitoring/grafana-dashboard.json
+++ b/charts/spegel/monitoring/grafana-dashboard.json
@@ -9,7 +9,6 @@
       "pluginName": "Prometheus"
     }
   ],
-  "__elements": {},
   "__requires": [
     {
       "type": "grafana",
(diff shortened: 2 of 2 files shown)


phillebaba:
Thanks for the contribution @AhmedTremo.

@onedr0p just wanted to check to see if you have any opinions about these changes.

onedr0p:
@AhmedTremo 

I don't see `spegel_mirror_requests_total` or `spegel_resolve_duration_seconds_bucket` being exported. I also don't have any data graphed in the `http_response_size_bytes_bucket` panels.

Are you on the latest version of spegel?

AhmedTremo:
@onedr0p 

I re-checked [pkg/metrics/metrics.go](https://github.com/spegel-org/spegel/blob/main/pkg/metrics/metrics.go) and these metrics are implemented and should be exportable. Are you sure Prometheus is correctly configured?

I the images I shared were for Spegel running v0.0.23 but I don't there were any metrics changes since then, (correct if I'm wrong @phillebaba)

onedr0p:
As I said it's probably on my end, but I do have other spegel metrics being ingested by prometheus.

AhmedTremo:
Yes, I think it's worth investigating. one thing to mention is I set 30-second window for calculating rate which might be shorter than scraping interval for some folks, and I do it for the metrics you mentioned

`spegel_mirror_requests_total`
`spegel_resolve_duration_seconds_bucket`
`http_request_duration_seconds_bucket`
`http_response_size_bytes_bucket`

So, I will update it to 5m just to be safe. Can you check if updating the window shows the metrics for you? @onedr0p

onedr0p:
Now I am getting data in the `http_request_duration_seconds_bucket` and `http_response_size_bytes_bucket` panels. Still no dice for the other two.

AhmedTremo:
I think we would someone else to try it and validate to be sure

phillebaba:
Partly my fault but this PR has been sitting for too long so the dashboard has changed a lot since it was created. I need to setup a proper environment myself where I can develop and test future changes for the dashboard, as there are a lot of improvements that I want to do in the future.

Closing this PR to keep open PRs clean.

PR #613, opened 2024-10-20
update default registries to include truecharts and Talos

PrivatePuffin:
Both projects use their own image store for, often repeatedly reused, containers


phillebaba:
Could you add an entry to the changelog?

phillebaba:
Closing because stale.

PR #614, opened 2024-10-22
fix grafana dashboard

spagno:
Hi,
when you load the dashboad in the UI, the __input makes grafana asks for the datasource and the dashboard works. When you load the dashboard from grafana' sidecar, it doesn't work because DS_PROMETHEUS is not inizialized properly, so I added the variable 

I installed the chart and the dashboard worked flawlessy

regards


onedr0p:
This PR can be closed, it was fixed in https://github.com/spegel-org/spegel/pull/633

phillebaba:
Closing at it has already been fixed.

PR #621, opened 2024-10-30
docs: add technical flow mermaid diagrams

AhmedTremo:
## Motivation
I faced some issues when I used Spegel because of wrong assumptions that I've made on how the project works, which motivated me to dive into the code to understand better. So, I created these diagrams to help me understand and to help others who get to same situations I was in. 

## Diagrams
  - Core components architecture
  - Image pull flow
  - P2P interactions
  - State tracking / content advertisements.

## Caveat
(These diagrams are based on my current understanding of the inner workings of the project), I created the diagrams as mermaid diagrams to they can be corrected or updated in the future as the project evolves.


phillebaba:
First of all this looks great. It is really detailed and very informative! :rose: 

One small things. In the first diagram it shows that Spegel handles fallback. In reality it is Containerd that deals with the fallback to the original registry, Spegel has no knowledge of this.

AhmedTremo:
@phillebaba That's a good catch, I updated the diagram/description. Let me know if you have comments/suggestions for others.

PR #659, opened 2024-12-05
Bump go-libp2p to address quic-go CVE

dereknola:
Instead of https://github.com/spegel-org/spegel/pull/658, bump the correct `direct` dependency which uses the newer version of quic-go.


phillebaba (approved):
Thanks this is a better option.

Issue #687, opened 2025-01-11
Hardcoded cluster.local in DNS bootstrap causes incompatibility with custom cluster domain names


subnet-dev:
### Spegel version

v0.0.29

### Kubernetes distribution

Kubernetes Vanilla

### Kubernetes version

v1.30.8

### CNI

Cilium

### Describe the bug

**Description:**

The current implementation of Spegel's `bootstrap-kind=dns` uses a hardcoded domain name that includes `cluster.local`. This causes compatibility issues for clusters configured with a custom DNS domain.

**Error logs**
```log
Defaulted container "registry" out of: registry, configuration (init)
{"time":"2025-01-11T16:09:07.288810726Z","level":"INFO","source":{"function":"main.registryCommand","file":"/build/main.go","line":216},"msg":"running Spegel","registry":":5000","router":":5001"}
{"time":"2025-01-11T16:09:07.289012556Z","level":"INFO","source":{"function":"github.com/spegel-org/spegel/pkg/routing.(*P2PRouter).Run","file":"/build/pkg/routing/p2p.go","line":112},"msg":"starting p2p router","logger":"p2p","id":"/ip4/172.16.0.243/tcp/5001/p2p/12D3KooWR71bYFGuW7cwujnHATM6FHkKKwuFLrTHxAfeoHSPTYPN"}
{"time":"2025-01-11T16:09:07.289024484Z","level":"INFO","source":{"function":"github.com/spegel-org/spegel/pkg/state.Track","file":"/build/pkg/state/state.go","line":34},"msg":"running scheduled image state update"}
{"time":"2025-01-11T16:09:07.290876411Z","level":"ERROR","source":{"function":"github.com/spegel-org/spegel/pkg/routing.bootstrapFunc.func1","file":"/build/pkg/routing/p2p.go","line":215},"msg":"could not get bootstrap addresses","logger":"p2p","err":"lookup spegel-bootstrap.spegel.svc.cluster.local. on 172.18.0.10:53: no such host"}
{"time":"2025-01-11T16:09:07.29204247Z","level":"ERROR","source":{"function":"github.com/spegel-org/spegel/pkg/routing.bootstrapFunc.func1","file":"/build/pkg/routing/p2p.go","line":215},"msg":"could not get bootstrap addresses","logger":"p2p","err":"lookup spegel-bootstrap.spegel.svc.cluster.local. on 172.18.0.10:53: no such host"}
```

**Affected Code Snippet:**

[...]

PR #688, opened 2025-01-11
Remove hardcoded cluster.local from DNS bootstrap domain on helm chart

subnet-dev:
Tested with `helm template` and `helm lint`

Fixes: #687


phillebaba:
Thanks for the suggested solution. I solved this by making the domain configurable in #689.

PR #693, opened 2025-01-13
add commonLabels for pods

jurim76:
Allow to identify pods by commonLabels


phillebaba:
@jurim76 looks good, could you add an entry to the Changelog and we should be able to merge.

phillebaba (approved):

PR #703, opened 2025-01-17
Fix p2p router close panic and add tests

phillebaba:
Graceful shutdown was not working due to a index out of bound panic when bootstrap was run during shutdown. This fixes the issue and adds tests for the p2p shutdown.

### DIFF
--- a/pkg/routing/p2p.go
+++ b/pkg/routing/p2p.go
@@ -116,11 +116,11 @@ func (r *P2PRouter) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	err = r.host.Close()
+	if err != nil {
+		return err
+	}
 	return nil
-}
-
-func (r *P2PRouter) Close() error {
-	return r.host.Close()
 }
 
 func (r *P2PRouter) Ready(ctx context.Context) (bool, error) {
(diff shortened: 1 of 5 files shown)


pkg/routing/p2p.go
```diff
@@ -116,13 +116,13 @@ func (r *P2PRouter) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	err = r.host.Close()
```
brandond, line 119: Can we add a `<- ctx.Done()` here?

brandond: I will also note that this causes the host not to get closed cleanly if bootstrapper.Run() returns an error.

pkg/routing/p2p.go
```diff
@@ -223,8 +223,13 @@ func bootstrapFunc(ctx context.Context, bootstrapper Bootstrapper, h host.Host)
 		bootstrapCtx, bootstrapCancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer bootstrapCancel()
 
+		// TODO (phillebaba): Consider if we should do a best effor bootstrap without host address.
+		hostAddrs := h.Addrs()
+		if len(hostAddrs) == 0 {
```
brandond, line 228: This also causes problems, as closing the host above causes h.Addrs() to return nil.

phillebaba:
@brandond sorry for the late reply. I have been trying to look through the code and document things more for those importing Spegel as a library. I added a comment to the bootstrap interface that I intend the Run function to be blocking. In my head it simplifies things even for those bootstrappers that do nothing during Run.

I should have waited with these changes to when I cut a v0.1.0 version to indicate the breaking changes. That is my intention for the next release and for the future to not break the API.

Issue #587, opened 2024-09-27
Remove predefined node taint after spegel pod finished initializing on the node


shandr:
### Describe the problem to be solved

On EKS, with Karpenter for nodes provisioning, in 99%, the workload pod starts before the spegel daemonset configures the node. This makes spegel useless in dynamic EKS + Karpenter environments. The suggested in the FAQ solution **nidhogg** is very unstable and not working as expected or not working at all. Additionally, with Karpenter, adding a taint that Karpenter is unaware of on the node can lead to a situation when Karpenter will create new nodes, because the pod doesn't have the required toleration.

### Proposed solution to the problem

Use the "cilium" [approach](https://docs.cilium.io/en/stable/installation/taints/). With Karpenter (or some other methods) add [startupTaints](https://karpenter.sh/docs/concepts/nodepools/#cilium-startup-taint) with some predefined name. When spegel initialized, it should remove the taint from the node.

phillebaba:
I wish this was a solution provided in Kubernetes. There have been a couple of solutions proposed and none of them have been accepted. I wish that every project did not need to implement their own solution to fix this challenge.

I am for creating a solution that would work for both Karpenter users and Cluster Autoscaler users short term. Long term we should consider a KEP to solve this problem in Kubernetes.

shandr:
With Cluster Autoscaler it's also possible to add some custom taint to Managed Nodegroups that spegel can remove after initialization. So it should work for both.

phillebaba:
Removing node taints would require introducing Kubernetes API dependencies and RBAC permissions to Spegel, which goes against its design goals of staying lightweight and stateless.

For those looking for this kind of automation, the feature has been added to the Prisma controller, Spegel’s enterprise offering.

zoispag:
I will revive this issue, pointing to https://node-readiness-controller.sigs.k8s.io/ (alpha) which I believes will solve this very problem, without relying on nidhogg.

phillebaba:
@zoispag I think this is a great idea as the controller will hopefully become a native feature eventually. The original Nidhogg repo is not being maintained anymore and NRC seems like it may be around.

If you have success getting something working it would be great to add it to the docs and replace Nidhogg as the recommended solution.  Going forward it could also be a resource which could be added to the Helm chart as an opt in?

zoispag:
@phillebaba once I'm back from my vacation, I will experiment and document a working solution. Can we reopen this issue?

zoispag:
[...]

PR #707, opened 2025-01-21
Option to remove a "not ready" taint at startup

vflaux:
Allow spegel to remove a taint at startup.
This allow to create nodes with a taint that prevent pods to be schedule. Spegel then remove the taint when ready.

fix https://github.com/spegel-org/spegel/issues/587

- [x] New code has tests where applicable
- [x] The change has been added to the [changelog](./CHANGELOG.md).
- [x] Documentation has been generated if applicable.
- [x] The unit tests pass.
- [x] Linter does not report any errors.
- [x] All end to end tests pass.


vflaux:
I'm not yet satisfied with the fact that the removal is done just after starting the server & shutdown routines without a real validation. The server could have stopped directly (listening to the socket may have failed) and the taint would still be removed.

phillebaba:
@vflaux I need to think about if we want to do this in Spegel. For me it feels weird to have Spegel be responsible for removing the taint from it's own node. While I know that Kubernetes does not support any way of doing this right now.

michielvha:
Hello, any update on this? We are using spegel and it would be a huge benefit if this functionality was included.

y-rabie:
I second that this will be helpful. This is already the way cilium, for example, works. It removes the taint after it's ready.

kw80:
Hi @phillebaba did you have time to analyze this PR? We are waiting on this improvement, because today we need to get lucky that spegel is active first so it starts copying files from existing nodes rather than downloading them externally over-and-over. The provided workaround is to use [taints](https://spegel.dev/docs/faq/#what-should-i-do-if-other-pods-are-scheduled-on-new-nodes-before-spegel), but maybe you see other issues with this PR.. Thank you!

phillebaba:
I have mixed feelings about this feature. One is that we are re-adding Kubernetes dependencies after we finally managed to remove them. I am not really a fan of the thought that each individual application needs to add this feature and be responsible for it.

While this isn't a no I do need to understand what the issue with using [nidhogg](https://github.com/pelotech/nidhogg) is, other than it being another dependency? Maybe this is something that can be discussed at the [community meeting](https://spegel.dev/project/community/#meeting) next week?

phillebaba:
Thanks a lot for the work you put into this PR. The feature is clearly useful, and I can see how automating taint management could help in some setups.

That said, I don’t think this is something we should add to Spegel. Kubernetes-specific behavior adds complexity and maintenance overhead, and it would require each instance to have permissions to modify nodes, which I’d prefer to avoid.

This kind of functionality is available in Prisma, a tracker for Spegel that I’m currently working on. It’s built for enterprise use cases, runs as a single instance, and already integrates with Kubernetes through a mutating webhook. Prisma is not open source and extends Spegel with functionality aimed at specific enterprise needs. If that’s of interest, I’m happy to discuss it further.

I’ll go ahead and close this PR, but feel free to open an issue if you’d like to discuss the topic more. Thanks again for the contribution.

PR #762, opened 2025-03-01
fix(metrics): use appropriate buckets for response size

thejoeejoee:
Default prometheus bucket are not covering the real response sizes, buckets less then 1.0 doesn't make sense.

Scaled to more reasonable scale 1kB to 50MB, up to discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#747](https://github.com/spegel-org/spegel/pull/747) Update Go to 1.23.6.
 - [#750](https://github.com/spegel-org/spegel/pull/750) Rename append mirrors to prepend existing.
 - [#373](https://github.com/spegel-org/spegel/pull/373) Apply mirror configuration on all registires by default.
+- [#762](https://github.com/spegel-org/spegel/pull/762) Set appropriate buckets for response size
 
 ### Deprecated
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -48,6 +48,8 @@ var (
 		Subsystem: "http",
 		Name:      "response_size_bytes",
 		Help:      "The size of the HTTP responses.",
+		// 1kB up to 2GB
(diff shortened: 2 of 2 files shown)


phillebaba:
@thejoeejoee this seems like a good change. We should probably adjust this for latency in a separate PR also. I was hoping native histograms would be stable by now but that does not seem to be the case.

I think 1kb could be a good lower bound. A lot of manifests are somewhere around 100-500 bytes. Layers however are a lot larger, sometimes in the GB range. So I think we need to increase the upper bound in the bucket to cover this. Some value above 1GB could be a good upper bound. We may need to use exponential buckets to get this kind of coverage in the full range.

```go
package main

import (
	"fmt"

	"github.com/prometheus/client_golang/prometheus"
)

func main() {
	buckets := prometheus.ExponentialBuckets(1024, 5, 10)
	fmt.Println(buckets)
}
```

Using this exponential bucket setting would give a range between 1Kb and 2Gb with the buckets `[1024 5120 25600 128000 640000 3.2e+06 1.6e+07 8e+07 4e+08 2e+09]`.

Does this sound reasonable?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#743](https://github.com/spegel-org/spegel/pull/743) Charts - removed metrics label from bootstrap service
 - [#748](https://github.com/spegel-org/spegel/pull/748) Fix topology annotation.
+- [#762](https://github.com/spegel-org/spegel/pull/762) Metrics - appropriate buckets for response size
```
phillebaba, line 29:
```suggestion
- [#762](https://github.com/spegel-org/spegel/pull/762) Set appropriate buckets for response size 
```

thejoeejoee:
@phillebaba fair point, we usually don't work with 2GB+ layers, but seems fine; changed to proposed 1kB-2GB

phillebaba (approved):
Thanks for the changes :rose: 

I am aware of a couple users of Spegel who are distributing multi GB images between nodes, which is why we need to cover those cases. Eventually I or someone else should revisit the metrics once native histograms becomes stable.

PR #823, opened 2025-04-14
Add topology aware routing

### LINKED ISSUE #669: Network topology awareness

When doing lookups in the DHT the first peer found will be returned. The query does not take into consideration the network topology of the Kuberentes cluster. Most cloud providers have multiple zones per region. Network traffic within the same zone is usually faster and cheaper compared to cross zone traffic. Spegel should prioritize peers in the same zone over other peers when possible.

There has been some interesting research already done on the topic about partitioning.

https://research.protocol.ai/publications/enriching-kademlia-by-partitioning/monteiro2022.pdf

We need to decide on the best solution to partition the DHT to avoid added latency for lookups.

phillebaba:
This change will add network topology aware routing. When enabled, peers in the same zone will be tried first before other peers. This will hopefully reduce cost and increase performance. This implementation is most likely not the most efficient implementation but is hopefully better than nothing. Until we can verify that there are no negative effects at scale it will be marked as experimental.

The feature works by exposing a endpoint within the p2p network which allows sharing of its zone. As peers are discovered the zone is checked and compared. If the zone is the same the peer will be immediately added to the channel. If the zones are different of if the zone could not be determined the peer will be added to a buffer list. This buffer list will only be written from if the desired peers within the same zone has not been found. Such an implementation means that peers in the same zone will always be attempted first.

There are two main downsides of this implementation. The first being that when content only exists in peers within different zones, the minimum latency until the peer is returned is the same as the maximum lookup latency, which by default is 20ms. The other downside is that more peers need to be discovered due to the nature of Kademlia. As Kademlia does random lookups and there are three different zones there is a 33% chance that a peer is within the same zone. If we want to return 3 peers at most we will need to have to discover more than 3 peers to overcome the probability that only one of them is in the same zone.

In the future this feature may be revised based on future discovery. It will have minimal impact on end users as the API exposed to configure the feature will not need to change. All end users need to worry about is that the zone name is passed to Spegel on startup.

Fixes #669

### DIFF
--- a/main.go
+++ b/main.go
@@ -55,6 +55,7 @@ type RegistryCmd struct {
 	ContainerdContentPath        string        `arg:"--containerd-content-path,env:CONTAINERD_CONTENT_PATH" default:"/var/lib/containerd/io.containerd.content.v1.content" help:"Path to Containerd content store"`
 	RouterAddr                   string        `arg:"--router-addr,env:ROUTER_ADDR,required" help:"address to serve router."`
 	RegistryAddr                 string        `arg:"--registry-addr,env:REGISTRY_ADDR,required" help:"address to server image registry."`
+	Zone                         string        `arg:"--zone,env:ZONE" help:"If set the router will prioritize peers within the same zone."`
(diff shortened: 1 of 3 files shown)


phillebaba:
One thought is that instead of using peerstore metadata it may be better to use an in memory LRU cache to avoid an ever growing store size. As nodes leave the cluster we want to remove known metadata as it will no longer be needed.

PR #832, opened 2025-04-16
Add delete hook to cleanup configuration from host when chart is uninstalled

### LINKED ISSUE #67: Clean up containerd mirror configuration

phillebaba:
This only took me two years but finally was able to come up with a solution that I don't hate. Its also a solution that has been right in front of me the whole time.

The cleanup solution uses Helm delete hooks to create a DaemonSet when the chart is uninstalled. These pods run the cleanup command which will clear the Containerd mirror configuration directory and restore any files from the backup directory. The solution would have ended there if Helm supported health checking DaemonSets as part as hooks, but it doesn't.

To enable Helm to remove the DaemonSet when the cleanup succeeds we need to create a waiting Pod. Without this Pod Helm would remove the DaemonSet as soon as it was created. With the Pod Helm will wait for the Pod to successfully complete running. Nothing will be deleted until that happens. We use this to our benefit, by having the wait Pod poll the cleanup readiness probe of all the cleanup pods. This probe is only running once the cleanup has succeeded. When the probes from all of the Pods are successful the wait Pod will exit successfully which in turn triggers Helm to remove all of the hook resources.

Fixes #67

### DIFF
--- a/charts/spegel/templates/post-delete-hook.yaml
+++ b/charts/spegel/templates/post-delete-hook.yaml
@@ -1,0 +1,106 @@
+{{- if .Values.spegel.containerdMirrorAdd }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "spegel.fullname" . }}-cleanup
+  namespace: {{ include "spegel.namespace" . }}
+  labels:
+    app.kubernetes.io/component: cleanup
+    {{- include "spegel.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: "post-delete"
+    helm.sh/hook-delete-policy: "before-hook-creation, hook-succeeded"
+    helm.sh/hook-weight: "0"
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: cleanup
+      {{- include "spegel.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: cleanup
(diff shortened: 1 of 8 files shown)


charts/spegel/templates/post-delete-hook.yaml
```diff
@@ -0,0 +1,85 @@
+{{- if .Values.spegel.containerdMirrorAdd }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "spegel.fullname" . }}-cleanup
+  namespace: {{ include "spegel.namespace" . }}
+  labels:
```
phillebaba, line 7: Resource labels should differ from actual Spegel labels.

Issue #863, opened 2025-05-06
Create data dir in volume instead of root filesystem


rouke-broersma:
### Describe the problem to be solved

The new /var/lib/spegel data dir is created in the root filesystem. Due to this readOnlyRootFilesystem in the security context can no longer be set to true.

### Proposed solution to the problem

 If the data dir would (optionally) live in a volume, readOnlyRootFilesystem can once again be set.

mungo312:
I'm not sure if this is really the sollution for the problem. I think there is a Bug in #850, as it says "For now keys will not be persisted between restarts when running Spegel in Kubernetes". 

From reading this, I think the data directory should not be used at the moment when running in kubernetes. From my point of view its more a bug then an enhancement.

phillebaba:
Yes this is a bug, the way forward is to fix so that the path is not used. Then set a good default security context including readOnlyRootFilesystem so that this does not happen again.

PR #882, opened 2025-05-15
Disable data persistence on Kubernetes

spiarh:
Fixes: #863


spiarh:
@phillebaba This disables the persistence by default but it is not clear to me if that would make sense/be better to enable it?

PR #943, opened 2025-07-17
Add spegel component label to spegel DaemonSet

bencoxford:
On bootstrap, spegel is trying to resolve the `spegel-cleanup` pods causing error in spegel logs so it fails to start up with a single node.


phillebaba:
@bc185174 thanks for fixing this. I don't know why I did not do this when I literally added a component label to the cleanup pods.

PR #975, opened 2025-08-13
Update Go to 1.25.0

phillebaba:
This updates the toolchain to use Go 1.25 when building.


phillebaba:
Need to wait for https://github.com/golangci/golangci-lint/issues/5873 to be fixed before merging.

phillebaba:
Applied fix suggested in https://github.com/golang/go/issues/75031 to get test coverage working. This needs to be revisited in the future.

Issue #545, opened 2024-07-19
Question: is it possible to push to Spegel?


caruccio:
### Describe the problem to be solved

Not a problem, only a question. Jjust found this project and could not find this question not in PRs, issues or docs.

Is it possible to push images to one of the Spegel instances and have it distributed to the whole cluster?
My use case is to avoid the pull from the first container at all, while pushing my new image right from the CI pipeline.
One simple solution for this is that I could just create a single pod only to trigger the pull, but it seams to be odd.

phillebaba:
So currently it is not possible but it is something that I have considered. It does however add some additional complexity to the solution. Spegel will always be a stateless application which means that all images stored are ephemeral. The risk then is that you build and push an image to one node, and when that node is removed from the cluster the image disappears. This could be mitigated with having image replication management like i mention in #375 but that is right now just an idea.

Is the reason you want to do this because you want to make sure the image is only pulled once? Or is it because you want to make sure the image is cached in the cluster before you start the deployment of a new application version?

caruccio:
Both, in fact. With a push model I can avoid the high latency of the first pull, while using high speed local network to replicate it over the rest of the cluster.

phillebaba:
Ok I can see that. I am not opposed to the idea, its just that I need to think through it a couple of times to come up with a good plan. This would add a write component to Spegel which would add additional complexity and security concerns. One thought that I had was to use a separate image store for images that have been pushed, then when another node pulls the image it will be in Containerd and we can remove it from the local store. Implementing this could work. I would probably have to change the configuration to a file to allow for some more complex settings. It would also enable serving content from multiple sources at once also, which could be an interesting feature.

srgvg:
I'd like to add a use case for being able to push images: a situation of air-gapped and offline installation, where no local registry is available.

Another solution for this use case could be https://zarf.dev/ which has an interesting solution to bootstraps a registry in the cluster. and then push images that are locally packaged (in a zarf archive). Whilst interesting this does not solve caching the controlplane and related components (k8s controllers, kubelet, pause container, ...). It also currently breaks my Gitops/FluxCD workflow. That latter also involves needing to cache other types of OCI artefacts (helm charts and FluxCD with yaml manifests).

vaizki:
Airgapped user here as well with no great solutions to bootstrap & run clusters with no registry connectivity...

phillebaba:
[...]

PR #977, opened 2025-08-14
Add support for push endpoints to registry server

wjordan:
This PR is an attempt at #545, making it possible to push to Spegel by implementing a subset of the OCI Distribution specification's endpoints related to pushing images in the registry server (the `/blobs/uploads/` and `PUT /manifests/` endpoints). The new endpoints are only handled when the `Push` config-flag is enabled (default `false`).

An optional `PushUpstream` extension (default `false`) asynchronously pushes uploaded images back to the upstream registry, and a temporary lease (`LeaseDuration`, default `10m`) is placed on newly created content and images to prevent them from being immediately garbage collected.

My use-case for this feature is to have Spegel act as a proxy for an upstream image push while making it immediately available in the local mirror, which can significantly improve latency and network usage compared to having the image perform a round-trip to/from the remote upstream registry.

---
This PR works well enough for my use-case at this point. I'd like feedback and guidance on whether this seems like a decent enough approach for being eventually merged into the main project with some further work.

The implementation currently uses the containerd `client.Client` directly and doesn't include any tests. The next step would probably be to abstract the implementation into corresponding `oci.Store` / `oci.Client` interfaces and add some tests against an in-memory implementation. Also open to any more specific (or entirely different) directions you'd like to go with this implementation as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/containerd/errdefs v1.0.0
 	github.com/containerd/typeurl/v2 v2.2.3
 	github.com/go-logr/logr v1.4.3
+	github.com/google/uuid v1.6.0
 	github.com/ipfs/go-cid v0.5.0
 	github.com/libp2p/go-libp2p v0.43.0
 	github.com/libp2p/go-libp2p-kad-dht v0.34.0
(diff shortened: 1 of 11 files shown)


phillebaba:
@wjordan thank you for the contribution! Adding the ability to push to Spegel has a lot of nuances that are not immediately. The main one is that it introduces a level of state fullness  as users pushing images to Spegel would expect them to be persisted. If the node the image was pushed to is removed for whatever reason that image would no longer be available. Additionally any image push would have to be done to a specific node to assure that the complete image has been pushed.

I have been doing some experimentation in Primsa, the enterprise offering for Spegel, to implement replication factors to allow instances to push layers in between each other to create a distributed stateful store. The reality is that the maintenance cost of such a feature does not make it feasible to add to Spegel. Especially because it becomes very difficult to debug during the initial development without access to the Kubernetest clusters. 

I am happy to discuss things further in Slack or during the community call to understand the specific problem you are trying to solve better, and maybe suggest alternative solutions to your problem.

phillebaba:
Closing as push is out of scope for the project goals.

PR #985, opened 2025-08-22
Add support for matching all registry hosts using '*' or '_default'

brandond:
Allow use of `*` or `_default` to match all registry hosts, for compatibility with containerd config.toml/certs.d.


phillebaba:
This code has been refactored slightly and now resides here. The constant string could be changed to a slice to make things simpler.

https://github.com/spegel-org/spegel/blob/8f30d7df91b4c07ee6fa69c9224aa51a7fdd27bb/pkg/oci/containerd.go#L625-L643

brandond:
Updated

Issue #656, opened 2024-12-04
Include the registry in the request log


davidhao3300:
### Describe the problem to be solved

Hello! Thank you for this great project. We'd like to track the cache misses by registry/repo in Datadog. We're able to get most of the way by using log-based monitoring using the log line logged [here](https://github.com/spegel-org/spegel/blob/0801ee9872d6dca1e92302a9d6a216a8a98dfbe1/pkg/registry/registry.go#L141). The main issue is that while we can parse the repository from the path field, we cannot figure out the registry since the "ns" URL param is not logged.

### Proposed solution to the problem

The ideal solution would be to parse the ns URL param and put a "registry" field in the log line, but I'm not sure if this is something that work for everyone. Perhaps putting an allow-listed set of URL params instead would be more-generic?

PR #986, opened 2025-08-22
Improve request logging

brandond:
* log request registry namespace if set
* pass server logr through request context
* log response stats at V(1)


phillebaba:
I like the idea of adding the logger to the request context. Makes things very clean and makes it possible to remove the logger from the registry. This way all HTTP servers in Spegel would take a logger when returning the handler instead. I can see why you would want to disable the logging  of the registry requests for k3s, as all the logs come as one. The issue right now is that there is no good way of determining if Spegel is working without the logs, but thats something I am working on.

Adding the ns parameter to the logs would fix #656, I would maybe just do this in a slightly different way. I will fix those things in a separate PR.

PR #987, opened 2025-08-22
Use TrackerOption to pass settings to Track

brandond:
Issue:
* https://github.com/spegel-org/spegel/issues/1023

Use TrackerOption to pass settings to Track and add new option to override the stats update interval, so that we can get image metrics that are updated more frequently than once every 9 minutes.

By default the tag/layer metrics are only updated once every 9 minutes (`routing.KeyTTL - time.Minute`). This is an excessively long time to go without updating stats, and frequently leads users to believe that spegel is not working. For example, if you start a new K3s node with an empty containerd image store, spegel metrics will show 0 tags and 0 layers for the first 9 minutes of uptime.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,12 +24,12 @@ require (
 	github.com/pelletier/go-toml/v2 v2.2.4
 	github.com/prometheus/client_golang v1.23.0
 	github.com/prometheus/common v0.65.0
-	github.com/stretchr/testify v1.10.0
+	github.com/stretchr/testify v1.11.1
 	go.etcd.io/bbolt v1.4.3
 	golang.org/x/sync v0.16.0
 	google.golang.org/grpc v1.75.0
-	k8s.io/apimachinery v0.33.4
-	k8s.io/cri-api v0.33.4
+	k8s.io/apimachinery v0.34.0
+	k8s.io/cri-api v0.34.0
 )
 
 require (
--- a/go.sum
+++ b/go.sum
@@ -463,8 +463,8 @@ github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
 github.com/stretchr/testify v1.8.3/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
 github.com/stretchr/testify v1.8.4/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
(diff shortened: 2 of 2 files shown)


phillebaba:
The changes with the configuration options are good. The question of the expiration ticker is a bit more nuanced than just updating metrics. The state tracker needs to re-advertise the store content before the TTL of the record written by the router. So decreasing the TTL in the state tracker will no only update the metrics faster but also advertise content more often, which in turn leads to a lot more network traffic in large clusters.

What we actually need is some other method of storing the advertised keys. Maybe it is a poor design choice to have the state tracker be responsible to re-advertising the keys before the TTL lapses. Instead this logic should be shifted to the p2p router itself. Long term something like that would also be beneficial to the p2p network.

Could we make the options changes and discuss the TTL questions in Slack/GH Issue/Community Meeting?

brandond:
My understanding of how libp2p works is that advertising content does not actually send anything out to the mesh. If another node searches for content with a given hash, the search is flooded across the network, until it reaches a node that can provide the content. Nodes do not actually store any information about which other nodes can provide content for a given hash.

In this case, refreshing content before the TTL is about to expire does not actually put any additional load on the network

Perhaps I am wrong?

phillebaba:
Under the hood it uses Kademlia for content discovery. In such a network both nodes and content have identifiers which are basically 256 bit shasums. Nodes in the network keep track of other nodes with a low XOR distance from its own identifier. When content is advertised the node will calculate the XOR distance between the content id and the node id, and "store" the content on those nodes. The nodes will in turn do the same until they are the node with the smallest XOR distance.

This is what makes the lookups so fast in very large networks. When you want to find content you ask the node with the smallest XOR distance between the content and node id, who will forward the query if it doesn't know where the content is.

So every advertisement has some network cost related to it. There are probably some smarter ways of dealing with this by having the router keep track of when the last advertisement was, as it is an implementation detail of libp2p. Some additional complexity exists here because we dont have content delete events from Containerd yet. Which means there is no way of removing content.

brandond:
> Some additional complexity exists here because we dont have content delete events from Containerd yet. Which means there is no way of removing content.

Last time I looked at it, there was no way of un-advertising things via libp2p either, or at least I couldn't find a function to do so. It looked like you just have to wait for the TTL to expire.

> Decreasing the TTL in the state tracker will no only update the metrics faster but also advertise content more often, which in turn leads to a lot more network traffic in large clusters.

Is there a way to query libp2p to see if a given hash is already advertised, and how long remains on the TTL? If so maybe we could only re-advertise if the remaining TTL is below a certain threshold?

The stats interval needs to be addressed sooner rather than later, we get a LOT of questions about how to tell if spegel is working and I'd love to have an answer other than "check debug logs".

phillebaba:
Changing the TTL will be possible once #1042 is implemented. I still don't think this would solve the issue referenced in the issue. I think the original cause of the issue was that content advertisement did not work properly which has now been fixed. Once #1042 has been implemented we can reopen this discussion in a separate issue.

PR #988, opened 2025-08-22
Public web fixes

brandond:
* make web public
  so that it can be used by other projects
* use WebOption to pass settings to Web, and add new options to pass transport and registry address
  allows running spegel with https
  allows running spegel on ports other than 5000
  allows running spegel with TLS client certificate auth
* handle errors reading metrics
  properly handles error responses from metrics endpoint
  fixes panic when metric families are not available (ie before tracker has synced tag/layer list)


brandond:
@phillebaba any progress on making the web stuff public so that we can expose it in embedding projects?

phillebaba:
Closing in favor of #1065

PR #989, opened 2025-08-22
Return context error when context is done

brandond:
Allows callers of state.Track to see error returned by context cancellation.


phillebaba:
Closing in favor of #996

PR #992, opened 2025-08-27
feat(helm): Add support for annotations on DaemonSet

NagaZuni:
## Summary
Adds configurable annotations to the Spegel DaemonSet via Helm chart.

## Changes
- Added `daemonsetAnnotations` field to values.yaml
- Updated DaemonSet template to include annotations
- Generated Helm documentation

## Usage
```yaml
daemonsetAnnotations:
  example.com/annotation: "value"
```
---
## Testing
Verified template rendering with test annotations in values.yaml using:
```bash
helm template ./ -f values.yaml

### DIFF
--- a/charts/spegel/values.yaml
+++ b/charts/spegel/values.yaml
@@ -25,7 +25,7 @@ serviceAccount:
   name: ""
 
 # -- Annotations to add to the DaemonSet.
-annotations: {}
+daemonsetAnnotations: {}
 
 # -- Annotations to add to the pod.
 podAnnotations: {}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: charts/spegel/values.yaml [hunk 1]
```diff
@@ -24,6 +24,9 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template.
   name: ""
 
+# -- Annotations to add to the DaemonSet.
+annotations: {}
```
phillebaba, line 28: Rename this tp `daemonsetAnnotations` to make it more clear.

phillebaba (changes requested):
I am not sure what the use case for adding annotations to the daemonset is, but I am fine with adding it if you can just rename the variable.

charts/spegel/values.yaml
NagaZuni: I have renamed the variable, and the annotations are primarily to let that DaemonSet be explicitly included or exempted from specific OPA Gatekeeper policies running in a cluster.

phillebaba (approved):

PR #1003, opened 2025-09-05
feat: add support for filtering registries and tags

### LINKED ISSUE #997: Ignore tags

Right now, the documentation about [tags](https://spegel.dev/docs/guides/updating-latest-tag/) provides the ability to ignore latest or tags.

Would be nice if spegel, instead of going from resolve everything to resolve nothing, provides an intermediate solution without using k8s-digester. 

There are projects that use other tags to keep an updated version of the project, like `stable` etc. which can not be resolved unless we deploy another component in the clusters like k8s-digester o
r decide to disable resolving all the tags.

instead of using the `resolveLatestTag: false` include an option `ignoreTags` which provides a list of tags or patterns that can be ignored by spegel.

option1:

```
spegel:
  ignoreTags: latest,1234-latest,stable,dev...
```
option2:

```
spegel:
  ignoreTags: *latest, stable
```

If the proposal is accepted, i'm willing to contribute to it

pepordev:
fixes #997

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ go.work
 
 # Added by goreleaser init:
 dist/
+
+# IDEs
+.vscode/
+.idea/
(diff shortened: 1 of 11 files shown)


phillebaba:
Thanks for the contribution, for the most part things look good. :rose: 

We are going to need to keep the original behavior of resolve latest tag for the time being so that it does not break existing use cases. I will deprecate the feature and remove it at a later time. It should run first before any registry filter.

I have some opinions about the general structure of the code to keep things simple. Instead of creating a new package called filter, I would compile all of the regex in main and then just pass a slice of regex. This is done for multiple reasons. Both to verify that the regex is valid, but also because there is a cost associated to parsing regex strings which we don't want to do for every request.

phillebaba:
This is looking good, could you add an entry to the changelog and then  rebase and squash into a single commit. Then we should be ready to merge.

phillebaba:
Looks like helm docs needs to be generated again.

pepordev:
> Looks like helm docs needs to be generated again.

should be fixed.

Issue #1014, opened 2025-09-16
Remove GOMAXPROCS env var.


uhthomas:
The GOMAXPROCS env var should no longer be necessary with Go 1.25.

https://go.dev/blog/container-aware-gomaxprocs

https://github.com/spegel-org/spegel/blob/d2c4340f174324bf98d03ab150f800a29c28ffec/charts/spegel/templates/daemonset.yaml#L119

PR #1015, opened 2025-09-16
chore: remove explicit GOMAXPROCS

uhthomas:
This is now done automatically in Go 1.25.

https://go.dev/blog/container-aware-gomaxprocs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [#1013](https://github.com/spegel-org/spegel/pull/1013) Refactor image parsing to include options.
+- [#1015](https://github.com/spegel-org/spegel/pull/1015) Remove explicit GOMAXPROCS
 
 ### Deprecated
 
--- a/charts/spegel/templates/daemonset.yaml
+++ b/charts/spegel/templates/daemonset.yaml
@@ -115,13 +115,6 @@ spec:
         env:
         - name: DATA_DIR
           value: ""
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/spegel-org/spegel
 
-go 1.24.0
-
-toolchain go1.25.1
```
phillebaba, line 5: We don't want to change this. The toolchain version tells Github actions and also your local go tooling which Go version to use when building Go, while the Go version indicates the minimum version required when importing Spegel as a library.  AS we are not using any Go 1.25 specific language features we want to keep the minimum version where it is as k3s import Spegel as a library.

phillebaba (changes requested):
Thanks for making this change, I thought there was something I forgot when I updated to Go 1.25. Could you fix my comment and also add an entry to the Changelog and we should be good to merge.

go.mod
uhthomas: This is done automatically by go mod tidy when changing the module version to 1.25.1, and the build won't succeed otherwise. The only way to prevent it is to have different versions for the module and the tool chain.

uhthomas: Happy to keep the module version as Go 1.24 if you think it's the right thing to do, though bear in mind there may be language specific changes between major versions which the toolchain directive won't apply. The most obvious example I can think of is with for loop scopes. 

https://go.dev/blog/loopvar-preview

I guess the GOMAXPROCS changes aren't a language change so doesn't apply here though.

uhthomas:
Fixes https://github.com/spegel-org/spegel/issues/1014

PR #1018, opened 2025-09-22
Add optional default values when parting image reference

phillebaba:
This change enables parse image to allow references missing the registry and optional library namespace. Strict parsing is still the default and removing the requirement for digests and appending default values have to opted in. We want to encourage proper references, but there are situations where we cant control that the registry is not set.

PR #1020, opened 2025-09-23
Don't filter ContentCreate events with no labels

brandond:
Fixes the issue with ContentCreate filtering preventing blobs from being advertised until the next tracker scan interval, as discussed at https://github.com/spegel-org/spegel/pull/1018#issuecomment-3321798056

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -444,7 +444,7 @@ func (c *Containerd) convertEvent(ctx context.Context, envelope events.Envelope)
 		if err != nil {
 			return nil, err
 		}
-		if !filter.Match(content.AdaptInfo(info)) {
+		if len(info.Labels) != 0 && !filter.Match(content.AdaptInfo(info)) {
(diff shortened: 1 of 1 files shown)


brandond:
My mistake, I think this was actually broken in https://github.com/spegel-org/spegel/pull/888 as we now only get OCIEvent for the image tag, not the digest, if the ContentEvent feature is enabled. The content filtering is an additional problem.

Honestly, I think the lack of ability to effectively filter content by image it is associated with probably makes ContentEvents a non-starter. Walking the image ensures that we only advertise content if it is associated with an image that matches the filter. On the backend, containerd doesn't care what images content is associated with - I'm not sure it even stores a reverse lookup for that; I think it just walks the whole image list and then deletes anything that's not referred to, every time it does a GC. And Spegel still walks the image every refresh interval, so any optimization associated with using content events instead of walking the image the first time we see it doesn't make a huge difference in performance.

PR #1031, opened 2025-10-01
Display errors in debug when pull measuring fails.

vflaux:
In the debug UI, errors like invalid image are not visible.
This PR allow to display these errors.

### DIFF
--- a/internal/web/web_test.go
+++ b/internal/web/web_test.go
@@ -1,9 +1,11 @@
 package web
 
 import (
+	"errors"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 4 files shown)


phillebaba (changes requested):
Exposing the error message in the debug view is a good idea. I would just ask to not reuse httpx.BasicResponseError for this. It was mostly created as a simple implementation to enable testing.

Instead please create a new Error struct in the web package called something like HTMLResponseError or similar. That does the required formatting needed.

phillebaba:
@vflaux could you also make an entry to the changelog?

internal/web/web_test.go
```diff
@@ -78,3 +80,52 @@ func TestDuration(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			e := NewHTMLResponseError(tt.err)
+			body, contentType, err := e.ResponseBody()
+			if tt.wantErr {
+				assert.Error(t, err)
```
phillebaba, line 123: Change these to require.Error

phillebaba: Require will error immediately while assert will fail the test but keep executing.

PR #1032, opened 2025-10-01
Pass registry address to debug web instead of using hard coded localhost

vflaux:
Debug UI use a hard-coded value (`localhost:5000`) for registry address instead of the one configured.
This PR use the value from configuration.

Also fixed in #988.


phillebaba:
@vflaux this looks good :rose: Could you just fix the field alignment lint error and add an entry to the changelog then we should be ready to merge.

PR #1033, opened 2025-10-02
Fix measuring pulls using digests

vflaux:
Measuring image pull with debug web UI with a digest in image results in failure (example: ghcr.io/spegel-org/spegel@sha256:a86089ae74c4f9c98ec86c366d196f7a03044c38af09e6582b0661d42a324226).

This PR fix this by using the parsed image and introducing the `Image.Reference()` method which is a copy of `DistributionPath.Reference()`.
This results in a similar call to `router.Resolve()` as in the `Registry` implementation.


phillebaba:
Thanks for finding and fixing my oversight. As with the other PRs just add a changelog entry and we should be good to go.

PR #1036, opened 2025-10-04
Remove is latest check in favor of registry filter

phillebaba:
This change removes the is latest check from both the state and registry. It leaves the flag which will instead add a filter. The flag will also be removed after the next release.

### DIFF
--- a/main.go
+++ b/main.go
@@ -191,7 +191,6 @@ func registryCommand(ctx context.Context, args *RegistryCmd) error {
 	// Registry
 	registryOpts := []registry.RegistryOption{
 		registry.WithRegistryFilters(args.RegistryFilters),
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -191,7 +191,6 @@ func registryCommand(ctx context.Context, args *RegistryCmd) error {
 	// Registry
 	registryOpts := []registry.RegistryOption{
 		registry.WithRegistryFilters(args.RegistryFilters),
-		registry.WithResolveRetries(args.MirrorResolveRetries),
```
brandond: Why did you remove this? This change does not seem to be related to removing the latest check, and this means the `--mirror-resolve-retries` flag is no longer honored.

PR #1046, opened 2025-10-14
feat: add GrafanaDashboard resource to Helm chart

onedr0p:
This adds the ability to use a `GrafanaDashboard` from the https://github.com/grafana/grafana-operator

I tried to make this a non-breaking change for people already using the dashboard, but I really struggled to find a decent pattern I was happy with for that. I think that since this is such a low impact breaking change this is fine.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -132,7 +132,7 @@ func NewRegistry(ociStore oci.Store, router routing.Router, opts ...RegistryOpti
 func (r *Registry) Handler(log logr.Logger) *httpx.ServeMux {
 	m := httpx.NewServeMux(log)
 	m.Handle("GET /readyz", r.readyHandler)
-	m.Handle("GET /livez", r.livenesHandler)
+	m.Handle("GET /livez", r.livenessHandler)
 	m.Handle("GET /v2/", r.registryHandler)
 	m.Handle("HEAD /v2/", r.registryHandler)
 	return m
@@ -153,7 +153,7 @@ func (r *Registry) readyHandler(rw httpx.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(http.StatusOK)
 }
 
-func (r *Registry) livenesHandler(rw httpx.ResponseWriter, req *http.Request) {
+func (r *Registry) livenessHandler(rw httpx.ResponseWriter, req *http.Request) {
 	rw.SetAttrs(HandlerAttrKey, "livez")
 
 	rw.WriteHeader(http.StatusOK)
@@ -306,7 +306,7 @@ func (r *Registry) mirrorHandler(rw httpx.ResponseWriter, req *http.Request, dis
 
 			rc, desc, err := r.ociClient.Fetch(fetchCtx, req.Method, dist, fetchOpts...)
 			if err != nil {
-				log.Error(err, "request to mirror failed, retryign with next")
+				log.Error(err, "request to mirror failed, retrying with next")
 				balancer.Remove(peer)
 				return false
 			}


charts/spegel/values.yaml
```diff
@@ -137,6 +137,18 @@ grafanaDashboard:
   # See: sidecar.dashboards.folderAnnotation for specifying the dashboard folder.
   # https://github.com/grafana/helm-charts/tree/main/charts/grafana
   annotations: {}
+  grafanaOperator:
```
onedr0p: Let me know if you think we should make this a top level `grafanaOperatorDashboard` and adjust the logic to suite it. I nested it for simplicity and they are related.

phillebaba:
I am a big fan of Grafana Operator, and I think it is something we should offer. I have however been thinking about how we package and deliver Grafana dashboards. Would it make more sense to create a separate repository and Helm chart with the Grafana dashboard? That would allow separate work on the dashboard that is not tied to the project. Would also make cutting releases for the dashboard easier.

My other thought is that some people may not want to add the dashboard in every cluster, but only in their cluster that runs Grafana. Right now it is not possible to add just the charts without deploying Spegel.

Another option is to keep everything in the same repository but we create two Helm charts. Add a new chart for observability, like alertmanager rules and other nice things.

Happy for any feedback about this. Any thoughts about same repository vs. separate repository?

nissessenap:
My 5 cents

For me it would be no impact as long as we have a seperation of alerts/dashboards from spegel it self.

It's probably easest to just put in a seperate repo and it's own helm chart.
The bad thing is of course that we probably want to have some compatibility matching between the spegel version and the monitoring helm chart.

Or keep it in the same repo, and have them match the current release flow which will slow down any alerts/dashboard updates.
Or update a bunch of release CI to support two different tags.

In general I'm anti mono-repo, so I would vote for it's own repo.

onedr0p:
> Would it make more sense to create a separate repository and Helm chart with the Grafana dashboard? That would allow separate work on the dashboard that is not tied to the project. Would also make cutting releases for the dashboard easier.

@phillebaba I don’t think that’s really necessary, especially for a project of this size and even bigger projects like Cilium, Fluent Bit, VictoriaMetrics, and the Grafana Operator include their dashboards directly in the same Helm chart. This approach feels more standard and avoids the extra complexity of managing a separate repo and/or chart just for observability.

> My other thought is that some people may not want to add the dashboard in every cluster, but only in their cluster that runs Grafana. Right now it is not possible to add just the charts without deploying Spegel.

That’s actually already possible. Since the dashboard isn’t templated, users can just use the [raw GitHub link](https://raw.githubusercontent.com/spegel-org/spegel/refs/heads/main/charts/spegel/monitoring/grafana-dashboard.json) directly. They can reference it in a `GrafanaDashboard` CR or include it with the Grafana Helm chart if they prefer. No need to deploy Spegel for that.

### Grafana Helm Chart

```yaml
dashboards:
  default:
    spegel:
      datasource: prometheus
      url: https://raw.githubusercontent.com/spegel-org/spegel/refs/heads/main/charts/spegel/monitoring/grafana-dashboard.json
```

### Grafana Operator

```yaml
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDashboard
metadata:
  name: spegel
spec:
  allowCrossNamespaceImport: true
  instanceSelector:
    matchLabels:
      grafana.internal/instance: grafana
  datasources:
    - datasourceName: prometheus
      inputName: DS_PROMETHEUS
  url: https://raw.githubusercontent.com/spegel-org/spegel/refs/heads/main/charts/spegel/monitoring/grafana-dashboard.json
```

Using those two methods would be no different if the dashboard file was elsewhere, including it in the helm chart is just a convenience so people don't need to use either of those methods and just use a 2 line config in the Spegel chart values.

phillebaba:
Ok I am convinced that we should just keep the dashboard in the Helm chart. I think that things looks good and the configuration is simple enough that it is easy for users to understand. I do appreciate you trying to keep things as similar as possible, and I do understand the need to break the label configuration. 

Long term I still think it would be interesting to add some alertmanager rules etc. to the chart or some other chart.

onedr0p:
@phillebaba thanks! I can look into contributing some PrometheusRules when I have a minute, I don't have any right now so any guidance on what would make sense would help.

PR #1054, opened 2025-10-24
Extended the debug web ui to provide more info

rutigs:
Added more support to the debug stats handler to provide more information to the UI including the current peers address, list of peers and their addresses and the available images on the current peer.

Screenshot 2025-10-23 at 10 41 07 PM

### DIFF
--- a/internal/web/templates/stats.html
+++ b/internal/web/templates/stats.html
@@ -2,7 +2,7 @@
   <div class="stat-container">
     <div class="stat-box" style="background-color: #64B5F6;">
       <div class="stat-title">Local Address</div>
-      <div class="stat-value" style="font-size: 18px; word-break: break-all;">{{ .LocalAddress }}</div>
+      <div class="stat-value" style="word-break: break-all;">{{ .LocalAddress }}</div>
     </div>
     <div class="stat-box" style="background-color: {{if .MirrorLastSuccess}}#81C784{{else}}#FAA93B{{end}};">
       <div class="stat-title">Last Mirror Success</div>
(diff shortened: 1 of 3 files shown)


internal/web/templates/stats.html
```diff
@@ -1,5 +1,9 @@
 <div>
   <div class="stat-container">
+    <div class="stat-box" style="background-color: #64B5F6;">
+      <div class="stat-title">Local Address</div>
+      <div class="stat-value" style="font-size: 18px; word-break: break-all;">{{ .LocalAddress }}</div>
```
phillebaba: Why set the value font size to be something different?

phillebaba (commented):
I think things look good, ran this in dev to see how things looked. The only comment I have is how we are passing the router and then casting the type. Personally I do not like how I read my own metrics in the current implementation. I had some thoughts about passing around a stats struct that would be populated by the different actors in Spegel.

As the debug view is showing things specific to the p2p router would it not be better to just pass the p2p router to debug web? Until some other method for sharing stats can be figured out.

phillebaba:
@rutigs could you also add a entry to the changelog?

phillebaba:
@rutigs if you fix my small NIT and then add a changelog I think we can merge this. I have some fixes to web view that I need to do anyway. I will end up changing the router to be P2P specific. Right now the memory router only exists for testing. We can treat the debug to be p2p specific until the hypothetical day that we implement a new kind of router.

rutigs:
> @rutigs if you fix my small NIT and then add a changelog I think we can merge this. I have some fixes to web view that I need to do anyway. I will end up changing the router to be P2P specific. Right now the memory router only exists for testing. We can treat the debug to be p2p specific until the hypothetical day that we implement a new kind of router.

Sounds good! Just gonna test the debug UI works the same with these changes right now and ill update the PR

PR #1059, opened 2025-10-26
main: return Verify() error

artem-tkachuk:
- Propagate `Verify()` failures instead of returning `nil` in the registry path


phillebaba:
You can remove the test cases that you added, and just fix the missing error return. Don't forget to add an entry to the changelog. 

The testing of main is something that has been in the back of my head for a while now, but it needs to be done properly with a running Containerd version. I have some ideas for how this will be done in the future but it is a bit more work to get nightly testing running with version matrixes.

artem-tkachuk:
Thank you for the review! Sounds great, I'll remove the test. Want me to take on the problem of testing main() too? Would be happy to sync up and/or brainstorm about this

artem-tkachuk:
@phillebaba done!

PR #1062, opened 2025-10-30
obs: add otel tracing

artem-tkachuk:
# Add OpenTelemetry tracing to Spegel

## Summary
- Add OpenTelemetry tracing to registry, metrics, and OCI HTTP flows
- Centralize HTTP instrumentation in `httpx` using `otelhttp`
- Respect existing OTEL providers/propagators when already configured
- Expose OTEL config in the Helm chart and wire flags in the DaemonSet
- Add unit tests covering OTEL setup, sampler behavior, and HTTP propagation

## Why
Provide request‑level tracing for debugging and observability without breaking embedders that already configure OTEL.

## Details
- Instrumentation:
  - HTTP handlers (registry, metrics) via `httpx.WrapHandler`
  - HTTP client transport (OCI requests) via `httpx.WrapTransport`
  - P2P operations (lookup/bootstrap)
- Helm: `values.yaml` exposes `.Values.spegel.otel.{endpoint,insecure,serviceName,sampler}` and flags are wired into the DaemonSet.
- OTEL setup avoids overriding global tracer provider/propagator when already set.
  
## Test plan
- `go test ./...`
- `golangci-lint run ./...`

## Impact
- OTEL is always built in; if a collector is not reachable, exporting may log warnings but does not fail requests.

## Screenshots
Screenshot 2026-01-22 at 19 25 57

Screenshot 2026-01-22 at 19 25 32

## Follow-up
- I am working on a separate PR to add the k8s itest tracing setup and document the repro steps.


internal/otelx/otelx_noop.go
```diff
@@ -0,0 +1,51 @@
+//go:build !otel
```
phillebaba: This should not be needed once OTEL is not an optional feature.

phillebaba (changes requested):
@artem-tkachuk as we discussed during last weeks community meeting. We should remove the optional build tags and then also the docker compose tests. Keep things slim with adding tracing, and then we can add more features as needed. Once you remove the things that are not needed it will reduce the noise and make it easier to review.

What would be nice is to get to a first point where we are setting up otelhttp so that both trace headers are forwarded by the clients and also registered for requests. Does that sound like a good idea?

artem-tkachuk:
@phillebaba thank you very much for reviewing! Will work on it today and tomorrow now that I have caught up after coming back from KubeCon :)

main.go
```diff
@@ -217,7 +250,7 @@ func registryCommand(ctx context.Context, args *RegistryCmd) error {
 	}
 	regSrv := &http.Server{
 		Addr:    args.RegistryAddr,
-		Handler: reg.Handler(log),
+		Handler: otelx.WrapHandler("registry", reg.Handler(log)),
```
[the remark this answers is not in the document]
brandond: Also keep in mind that embedding projects may have their own otel config that this will need to play nice with.

internal/otelx/otelx_noop.go
artem-tkachuk: Removed in 7b5cb69eedb317e99366e706dfc7085981791d67

main.go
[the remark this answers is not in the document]
artem-tkachuk: HTTP instrumentation now uses `otelhttp` and lives in `httpx` (handlers + transport), and OTEL setup skips overriding existing tracer providers/propagators if already configured 15df91fa380f16af03ebe45fb9caf8c382555745

artem-tkachuk:
Finished addressing feedback. Remaining items are adding more testing for codecov, providing otel endpoint for k8s itest, cleaning up commits, and updating description.

artem-tkachuk:
@phillebaba, could you add a label to the PR since they're now required?

artem-tkachuk:
Added more testing to improve codecov coverage, cleaned up commits, and updated the description. Ready for re-review!

main_test.go
```diff
@@ -0,0 +1,27 @@
+package main
```
phillebaba: Remove this file. I am working on getting coverage reports from the Kuberentes integrationtests to cover the main code.

charts/spegel/README.md
```diff
@@ -69,6 +69,11 @@ Read the [getting started](https://spegel.dev/docs/getting-started/) guide to de
 | spegel.mirrorResolveRetries | int | `3` | Max amount of mirrors to attempt. |
 | spegel.mirrorResolveTimeout | string | `"20ms"` | Max duration spent finding a mirror. |
 | spegel.mirroredRegistries | list | `[]` | Registries for which mirror configuration will be created. Empty means all registires will be mirrored. |
+| spegel.otel | object | `{"endpoint":"","insecure":false,"sampler":"parentbased_always_off","serviceName":"spegel"}` | OTEL tracing configuration. |
+| spegel.otel.endpoint | string | `""` | OTEL exporter endpoint (e.g., otel-collector:4318). |
```
svrnm: It's not clear if this supports both OTLP over grpc and http or only one. It's good practice to support both.

svrnm (commented):
Hi, I provided an initial feedback from OTel best practices perspective, I'll do some more review later but overall this looks like a good start.

pkg/httpx/otel.go
```diff
@@ -0,0 +1,25 @@
...
+
+// WrapHandler wraps an HTTP handler with OpenTelemetry instrumentation.
+func WrapHandler(name string, h http.Handler) http.Handler {
+	return otelhttp.NewHandler(h, name)
+}
+
+// WrapTransport wraps an HTTP transport with OpenTelemetry instrumentation.
+func WrapTransport(name string, rt http.RoundTripper) http.RoundTripper {
+	if rt == nil {
+		rt = http.DefaultTransport
```
phillebaba: This is already the default in NewTransport.

phillebaba (changes requested):

artem-tkachuk:
Addressed all the feedback in #1196 and merged #1197, now going to work on the feedback here. Thank you very much for reviewing @svrnm @brandond @phillebaba !

svrnm:
thank you @artem-tkachuk. Are you in the CNCF slack by any chance? I would like to ask a few people that know go better than I do to take a look as well, but it would be great if I can tag you in the conversation:-)

internal/otelx/otelx_otel.go
```diff
@@ -0,0 +1,168 @@
...
+		sdktrace.WithSampler(newSampler(cfg.Sampler)),
+	)
+
+	otel.SetTracerProvider(tp)
+
+	defaultPropagator := propagation.NewCompositeTextMapPropagator(
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	)
+	if reflect.DeepEqual(otel.GetTextMapPropagator(), defaultPropagator) {
```
dmathieu: Using reflect here is a bit strange. Are you expecting folks to set propagators in other places? Is that a good pattern?

pkg/httpx/otel.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+	if name == "" {
+		return otelhttp.NewTransport(rt)
+	}
+	return otelhttp.NewTransport(rt, otelhttp.WithSpanNameFormatter(func(operation string, _ *http.Request) string {
+		return name + " " + operation
```
dmathieu: I would recommend following the semantic convention for HTTP span names here.
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name

PR #1108, opened 2025-11-25
Break out of subCancel loop on context cancel error

brandond:
Also fixes capitalization and spelling - containerd branding consistently uses all lowercase for the name.


brandond:
@phillebaba the test failures here do not appear to be related, am I missing anything?

PR #1114, opened 2025-12-01
Fix empty json media type

erdody:
# Problem

Spegel doesn't work with [ORAS](https://oras.land/) OCI images.

If the content of a blob is an empty json ("{}"), which is the default behavior for the config layer in ORAS generated images (see [here](https://oras.land/docs/how_to_guides/manifest_config/#empty-config-file)), serving the blob fails with 
> could not determine media type

The media type is not correctly detected for `application/vnd.oci.empty.v1+json` blobs.

This was introduced in 0.5.0,  registry#blobHandler now [calls](https://github.com/spegel-org/spegel/blob/v0.5.0/pkg/registry/registry.go#L432) ociStore.Descriptor, which tries to infer the media type. That was not the case before.

# Solution

In `FingerprintMediaType()`, return `application/vnd.oci.empty.v1+json` if the content is an empty json.

### DIFF
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -73,6 +73,16 @@ func FingerprintMediaType(r io.Reader) (string, error) {
 		return "", errors.New("expected object start")
 	}
 
+	if !dec.More() {
+		b, err := io.ReadAll(dec.Buffered())
+		if err != nil {
+			return "", err
+		}
+		if len(b)+int(dec.InputOffset()) == 2 {
+			return ocispec.MediaTypeEmptyJSON, nil
+		}
+	}
+
 	schemaVersion := 0
 	mediaType := ""
 
--- a/pkg/oci/oci_test.go
+++ b/pkg/oci/oci_test.go
@@ -1,6 +1,7 @@
 package oci
 
 import (
+	"io"
 	"os"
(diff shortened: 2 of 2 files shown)


phillebaba:
@erdody sorry for the late review I was busy chasing down a difficult bug. Do you thing that checking the digest would be better than the fact that there are no keys in the map? I am not sure what constitutes emtpy json media type. Hoping to get this fixes before the next release.

erdody:
> @erdody sorry for the late review I was busy chasing down a difficult bug. Do you thing that checking the digest would be better than the fact that there are no keys in the map? I am not sure what constitutes emtpy json media type. Hoping to get this fixes before the next release.

Thanks for looking at it.

Looking at [the spec](https://specs.opencontainers.org/image-spec/manifest/#guidance-for-an-empty-descriptor), it seems like it's expected to always be "{}". I wanted to make it more generic and allow potential white spaces. Either one would work for me, lmk.

phillebaba:
Lets keep to the spec if it defines the empty descriptor content. So we should only return if the input matches the expected digest and byte size. Could you add a test for both `{}` and `{ }` where the first passes and the latter does not. Just so we are explicit with expected behavior in the tests.

PR #1145, opened 2025-12-16
Feature: Support SOCI Snapshotter

YQ-Wang:
This PR adds support for AWS SOCI (Seekable OCI) snapshotter, enabling P2P distribution of SOCI indices for lazy loading container images:

- Add SOCI media type recognition (V1 and V2)
- Classify SOCI indices as manifests in IsManifestsMediatype() to enable P2P distribution

The SOCI snapshotter (standalone daemon mode) does not support config_path to read containerd's hosts.toml files. Users must explicitly configure mirrors per registry.

Add to /etc/soci-snapshotter-grpc/config.toml:
```
[resolver.host."docker.io"]
  [[resolver.host."docker.io".mirrors]]
    host = "http://<NODE_IP>:<SPEGEL_PORT>"  # Same as in hosts.toml
    insecure = true

[resolver.host."ghcr.io"]
  [[resolver.host."ghcr.io".mirrors]]
    host = "http://<NODE_IP>:<SPEGEL_PORT>"
    insecure = true
```


brandond:
> AWS SOCI (Seekable OCI) snapshotter, enabling P2P distribution of SOCI indices for lazy loading container images

I've not heard of this before, how does this compare with https://github.com/containerd/stargz-snapshotter?

brandond:
It's unfortunate that AWS decided to reinvent their own lazy-loading container image format instead of just supporting stargz, which has been around since 2020.

han-steve:
Yeah I think they forked `estargz` since it requires all images to be in a certain format. We use AWS SOCI because it allows us to only add an index to some images on the node and leave other images as-is.

phillebaba:
@han-steve this used to be the case until a few months ago. The new SOCI format has reverted back to mutating the whole image instead of using the referrers API to get the index. I liked SOCI for the same reason you do so I think it's a shame that they changed this.

phillebaba:
@YQ-Wang looked over your PR now and have some questions about the changes.

A nuance with SOCI mirror configuration is that it is not hot reloaded, instead it is only loaded when SOCI is restarted. This makes things a bit tricky especially when wildcard registries are not supported.

The other part is that SOCI stores layers in a different format in its own directory. Containerd will still manage the manifests and then offload the fetching of layers to the snapshotters. So to fully be able to support SOCI we would need to access that data as well.

Just to make sure, was the intent to support only SOCI manifests or also the layers which SOCI stores?

Issue #1120, opened 2025-12-04
Constant `bootstrap completed but connectivity has not been reached` error spew on single-node cluster


brandond:
### Spegel version

main

### Kubernetes distribution

n/a

### Kubernetes version

n/a

### CNI

n/a

### Describe the bug

On single-node clusters, spegel continuously logs ERROR-level messages about connectivity problems every few seconds. The errors go away when another node joins the cluster, but this is very irritating.

> ```
> ERRO[0049] failed to run bootstrap                       attempts=17 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0057] failed to run bootstrap                       attempts=18 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0062] failed to run bootstrap                       attempts=19 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0071] failed to run bootstrap                       attempts=20 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0078] failed to run bootstrap                       attempts=21 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0080] failed to run bootstrap                       attempts=22 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0088] failed to run bootstrap                       attempts=23 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0096] failed to run bootstrap                       attempts=24 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0102] failed to run bootstrap                       attempts=25 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0109] failed to run bootstrap                       attempts=26 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0113] failed to run bootstrap                       attempts=27 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0114] failed to run bootstrap                       attempts=28 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0115] failed to run bootstrap                       attempts=29 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
> ERRO[0123] failed to run bootstrap                       attempts=30 error="bootstrap completed but connectivity has not been reached" logger=spegel/p2p
[...]

PR #1149, opened 2025-12-18
Allow disabling error-level logging of bootstrap connectivity failures

brandond:
Fixes:
* https://github.com/spegel-org/spegel/issues/1120

### DIFF
--- a/pkg/routing/p2p.go
+++ b/pkg/routing/p2p.go
@@ -53,6 +53,7 @@ type P2PRouterConfig struct {
 	Libp2pOpts        []libp2p.Option
 	AdvertiseTTL      time.Duration
 	MaxReprovideDelay time.Duration
+	LogBootstrapErr   bool
 }
 
 type P2PRouterOption = option.Option[P2PRouterConfig]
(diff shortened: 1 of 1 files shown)


brandond:
cc @phillebaba - any chance of getting this looked at?

Should we wire this up to a CLI flag?

phillebaba:
I am currently taking some time off from working on Spegel until the beginning on February, so that I can focus on work that will allow me to pay the bills. Until then I will not be active in issues or reviewing any PRs.

astro-stan:
hello, is there any progress on this?

brandond:
@phillebaba should I bother updating this PR - are you planning on accepting this in any form, or should I just keep it in the K3s fork?

phillebaba:
In what situation is someone running a single node k3s cluster with Spegel enabled? There is no reason for it as there is no peer to pull image from. Just trying to understand the use case for disabling the error.

brandond:
Users frequently configure their clusters identically regardless of their size. 

Since servers and agents can be added and removed ad-hoc, it is not required that a user decide ahead of time if they will add additional nodes to a cluster. It is completely reasonable that they might enable Spegel by default so that it is already configured if/when they eventually add more nodes.

Users may also just have a standard config template that they use for clusters of all sizes. There is no reason to penalize these users with a spammy log message on single-node clusters.

PR #1250, opened 2026-03-20
Extend HTTP bootstrapper to accept external URL

dimitar-kostadinov:
Spegel currently lacks a native bootstrap mechanism when running directly on a host, unlike Kubernetes deployments which rely on DNS bootstrapper.

This PR extends the HTTP bootstrapper so it can also be pointed at an external system that returns a set of bootstrap peers, enabling peer discovery in non-Kubernetes environments. It also includes optional mTLS support for secure communication with the external system.

The following flags are added:
- `--http-bootstrap-url` accepts a full URL of an external bootstrap endpoint (mutually exclusive with `--http-bootstrap-peer`, which keeps the existing peer-base-URL).
- `--http-bootstrap-ca`, `--http-bootstrap-tls-crt`, `--http-bootstrap-tls-key` add optional CA verification and mTLS for the bootstrap endpoint.

If `--http-bootstrap-addr` is omitted serving own `peer.AddrInfo` is disabled.

The HTTP bootstrap response format is a JSON array of `peer.AddrInfo` objects (the `ID` field may be omitted/empty).

Related to #829

### DIFF
--- a/pkg/routing/bootstrap.go
+++ b/pkg/routing/bootstrap.go
@@ -213,22 +213,29 @@ type ExternalBootstrapper struct {
 
 func NewExternalBootstrapper(url url.URL, tlsCA, tlsCert, tlsKey []byte) (*ExternalBootstrapper, error) {
 	client := httpx.BaseClient()
+	var tlsConfig *tls.Config
 	if len(tlsCA) > 0 {
-		tlsConfig := &tls.Config{
-			MinVersion: tls.VersionTLS12,
-		}
+		tlsConfig = &tls.Config{MinVersion: tls.VersionTLS12}
 		caCertPool := x509.NewCertPool()
 		if !caCertPool.AppendCertsFromPEM(tlsCA) {
 			return nil, errors.New("failed to parse CA certificate")
 		}
 		tlsConfig.RootCAs = caCertPool
-		if len(tlsCert) > 0 && len(tlsKey) > 0 {
-			cert, err := tls.X509KeyPair(tlsCert, tlsKey)
-			if err != nil {
-				return nil, fmt.Errorf("failed to parse client certificate: %w", err)
-			}
-			tlsConfig.Certificates = []tls.Certificate{cert}
+	}
+	if len(tlsCert) > 0 || len(tlsKey) > 0 {
+		if len(tlsCert) == 0 || len(tlsKey) == 0 {
+			return nil, errors.New("both client TLS certificate and client TLS key must be provided")
 		}
+		cert, err := tls.X509KeyPair(tlsCert, tlsKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse client certificate: %w", err)
+		}
+		if tlsConfig == nil {
+			tlsConfig = &tls.Config{MinVersion: tls.VersionTLS12}
+		}
+		tlsConfig.Certificates = []tls.Certificate{cert}
+	}
+	if tlsConfig != nil {
 		transport := httpx.BaseTransport()
 		transport.TLSClientConfig = tlsConfig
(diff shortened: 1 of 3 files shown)


phillebaba:
@dimitar-kostadinov sorry for the delayed review. I have been moving so all my time after work has been focused on getting settled.

This looks like a way forward to make non Kubernets deployments easier. What is the difference between this and the HTTP bootstrapper? I have considered looking over the bootstrappers that currently exist to see what we want to keep. K3S has their own bootstrapper that uses master nodes as the rendevouz. 

How would the other end of the server look like? Maybe it is better to discuss the design in a feature request first. Think it is good we are giving attention to the bootstrappers I just dont want to introduce another one that has overlap with something that already exists.

dimitar-kostadinov:
@phillebaba Thanks for taking the time to review this PR.

> What is the difference between this and the HTTP bootstrapper?

The main difference is that the External bootstrapper is intended to call a server that may live outside of the Kubernetes cluster. For that reason, it supports optional mTLS to secure the communication with the server.
The server also aims to return a set of bootstrap peers ([>2 if possible](https://github.com/libp2p/go-libp2p-kad-dht/blob/0ad6ca5eeecff43283a334120a12f8c0add79f1b/dht.go#L530-L538)), similar to the DNS bootstrapper.
The contract is simple: the server must return a `[]netip.Addr` array of bootstrap peers.

> How would the other end of the server look like? Maybe it is better to discuss the design in a feature request first.

The server implementation depends on the scenario, so multiple implementations are possible. For example, when the control plane lives outside of the Kubernetes cluster, one possible implementation is [this one](https://github.com/dimitar-kostadinov/gardener-extension-registry-cache/blob/spegel_api_poc/cmd/gardener-extension-spegel-bootstrap/main.go).

phillebaba:
Could we just merge the external and HTTP bootstrapper into one? Add the mTLS resources. Set an API for what the bootstrapper should return etc. Then just make it optional if the http bootstrapper should also serve its own identity back.

I also thing that a peer.ID should be returned instead of just the IP. This gives the option to return both the IP or the IP and public key.

pkg/routing/bootstrap.go
```diff
@@ -191,10 +249,45 @@ func (bs *HTTPBootstrapper) Get(ctx context.Context) ([]peer.AddrInfo, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	return addrInfos, nil
+}
+
+// bootstrapPeerAddrInfo mirrors libp2p's peer.AddrInfo JSON shape but allows the ID to be omitted or empty.
+// libp2p's peer.AddrInfo.UnmarshalJSON rejects an empty ID, so we unmarshal into this struct and convert via fromBootstrapPeerAddrInfos.
+type bootstrapPeerAddrInfo struct {
+	ID    string
```
phillebaba: NIT: Can we add explicit json annotations even if it works without.

pkg/routing/bootstrap.go
```diff
@@ -128,22 +133,74 @@ func (b *DNSBootstrapper) Get(ctx context.Context) ([]peer.AddrInfo, error) {
...
 	httpClient *http.Client
 	addr       string
-	peer       string
+	endpoint   string
 }
 
-func NewHTTPBootstrapper(addr, peer string) *HTTPBootstrapper {
+// NewHTTPBootstrapper creates an HTTP bootstrapper.
+//
+// Self-serving identity at /id is enabled iff addr is non-empty.
```
phillebaba: NIT: enabled "if"

phillebaba (commented):
Thank you for merging this with the HTTP bootstrapper. Two small NITs and then we can merge.

Issue #1096, opened 2025-11-18
Spegel does not advertise content loaded using `ctr image import`


brandond:
### Spegel version

All since #888

### Kubernetes distribution

k3s, rke2

### Kubernetes version

n/a

### CNI

n/a

### Describe the bug

* When spegel gets a ContentCreate event from containerd, it does not advertise the layer content if the event lacks a Digest:
https://github.com/spegel-org/spegel/blob/decaf73f5173215c5869d27b5fabbc3fb632e578/pkg/oci/containerd.go#L369-L372
  This is a problem, as the ImageCreate events generated by containerd when an image is imported from tarball only contain the tag, not the digest. There is no event generated for the image digest, so the image digest and layer content digests are never advertised.
* When containerd imports an image instead of pulling it, it does not add `containerd.io/distribution.source.<registry>=<name>` labels to the content. These labels are only added by the `docker` remote. This prevents the content from being properly tracked by spegel, as it applies label-based filters to image content:
https://github.com/spegel-org/spegel/blob/decaf73f5173215c5869d27b5fabbc3fb632e578/pkg/oci/containerd.go#L561-L578

Combined, these two issues make it so that spegel fails to advertise everything except the tag when content is imported instead of pulled.

brandond:
Related: 
* https://github.com/containerd/containerd/issues/8483

phillebaba:
Containerd seems to behave very differently when images are pulled with CRI vs other methods. It is hard to cover every single use case, when labels are applied differently. I do think that the addition of integration tests for different Containerd versions in #1097 is going to help a lot. Not only have I been dealing with different Containerd versions, but also different ways images makes its way in.

I am fine with fixing this as long as it can be covered in the integration tetsts.

brandond:
Yeah, unfortunately I don't see a way around this other than going back to the old way of not ignoring tags, and always walking images to find content digests. A bunch of the stuff that you have come to rely on for efficient discovery of resources to advertise is managed by the CRI service.

brandond:
FYI, we are currently planning to work around this with additional code on the K3s side to manually create digests and label layer content - by walking the image - when K3s imports images from tarballs. But that will not help anyone that uses `ctr` or `nerdctl` to import content.

phillebaba:
[...]

PR #1252, opened 2026-03-23
Add option to disable content events in Containerd store

farazkhawaja:
Workaround for issue https://github.com/spegel-org/spegel/issues/1096 https://github.com/k3s-io/k3s/issues/13794

- Adds a `WithContentEventsDisabled()` option to the Containerd store that disables reliance on ContentCreate events for advertising image content        
             
- When disabled, content is advertised via `ImageCreate` events by walking the image manifest tree, instead of waiting for individual `ContentCreate` events

- To be used in the following manner 
```
storeOpts := []oci.ContainerdOption{
		oci.WithContentPath(filepath.Join(nodeConfig.Containerd.Root, "io.containerd.content.v1.content")),
		oci.WithContentEventsDisabled(),
	}
```

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -43,7 +43,8 @@ const (
 )
 
 type ContainerdConfig struct {
-	ContentPath string
+	ContentPath          string
+	DisableContentEvents bool
 }
 
 type ContainerdOption = option.Option[ContainerdConfig]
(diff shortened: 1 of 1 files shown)


breboulet:
Is there anything that could be done to help this get merged in? We love spegel and this issue is preventing us from using the latest k3s versions unfortunately :(
BTW thanks @farazkhawaja for proposing a fix.

phillebaba:
I told  @farazkhawaja in Slack a while back to create an issue first before creating the PR so that the implementation could be discussed first. I do not think reverting content advertisement is the solution. 

Any solution needs integration tests to verify that imported images with ctr works with Spegel to avoid regression. The fact that Spegel supported this before was circumstantial rather than on purpose. Without integration tests this can never be guaranteed to work in the future. 

I have limited time right now which I am spending on fixing hedged requests. I am happy to discuss a support contract, if Spegel is a critical component at your work @breboulet and you need this fixed with an ETA.

farazkhawaja:
@phillebaba I'll create a new issue for this

PR #1258, opened 2026-03-31
 Hedge requests to peers to protect against tail latency

phillebaba:
This changes how we make requests to peer when mirroring a requests to protect against tail latency. The main reason large differences in latency occurs is mostly due to peers being removed but still being present in the DHT.

The logic that is implemented is meant to minimize latency while also reducing the amount of in flight requests. In the normal flow without hedges a request is immediately fired and the result is picked up in the result channel. Content is written back to the response and returned.

If the request takes too long to return a series of hedge requests will be triggered to race the original request. The first request to return is used to write to the response. When the response is being written to no hedge requests will be triggered. Response durations are measured and recorded with the hedger to determine the time to wait before hedging. There will be 3 hedge requests sent at p80, p90 and p95.

When a request dial succeeds but the actual request fails due to 404 or some other http status code the next request will be fired off immediately.

There are two failure conditions separate from the request timing out. These are only active when there are no in flight requests. The first is an idle timeout that sets the max amount of time doing nothing. The reason is that  the query can take a lot longer than expected to determine that no peer is found. To protect against this we chose to just assume no more peers will be returned after a set amount of idle. The second is a failure state when the query completes before the timeout and does not have any peers left. In this scenario the DHT has completed the walk and determined no peers can be found. As stated the failure conditions are only active when no requests are running to allow for all requests to complete.

Fixes #1193 
Closes #1195

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.26.0
 
 require (
 	cuelabs.dev/go/oci/ociregistry v0.0.0-20250530080122-d0efc28a5723
+	github.com/HdrHistogram/hdrhistogram-go v1.2.0
 	github.com/alexflint/go-arg v1.6.1
 	github.com/containerd/containerd/api v1.10.0
 	github.com/containerd/containerd/v2 v2.2.2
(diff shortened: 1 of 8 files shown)


et304383:
@phillebaba are you hopeful that this change will resolve issues we are seeing with timeouts all over the place when we have high node churn such as Karpenter consolidation?

Do you have an ETA for when you will be able to cut a new release and helm chart update?  I was evaluating Spegel last week and cannot possibly rollout to all of our environments without a fix for poor performance/timeouts during node replacement cycles such as rolling node replacement for AMI updates.

phillebaba:
Not at this time @et304383 ! Even though Spegel is used in multiple production grade environments, I do not focus on production grade support. I focus purely on the home lab or individual user use case. I wish you all the best with your evaluation.

et304383:
> Not at this time @et304383 ! Even though Spegel is used in multiple production grade environments, I do not focus on production grade support. I focus purely on the home lab or individual user use case. I wish you all the best with your evaluation.

I am confused.  You closed https://github.com/spegel-org/spegel/issues/1193 which I thought was the issue covering problems with high node churn.  Is this PR related or not?  I'm not asking for your guarantee this will work in prod, just if this update will improve things in those scenarios.

And again, do you have time to cut a new release and helm chart update?

PR #1277, opened 2026-04-10
Refactor balancer to become iterator whith better waiting logic

phillebaba:
This change splits out the changes in #1258 to make testing easier. The new iterator allows waiting with channels and makes acquire non blocking.


pkg/routing/p2p.go
```diff
@@ -279,39 +279,27 @@ func (r *P2PRouter) Ready(ctx context.Context) (bool, error) {
...
-			// Don't refresh if min count is already met.
-			if count > 0 && cb.Size() >= count {
-				cb.Close()
-				return cb, nil
-			}
-
-			// If we are running a refresh query we ant a new closer.
-			cb = NewClosableBalancer(cb.Balancer)
-			r.balancerCache.Add(c.String(), cb)
+			// TODO: Add condition to run update.
```
phillebaba: Reminder to fix this.

PR #1309, opened 2026-04-21
fix: update Grafana dashboard datasource configuration for Prometheus

### LINKED ISSUE #1303: Grafana Dashboard broken for Victoria Metrics
### Spegel version

0.7.0

### Kubernetes distribution

Talos

### Kubernetes version

1.35.4

### CNI

Cilium

### Describe the bug

While 03e48c98de709b5d3c9893aa41b52604a921f208 fixed the dashboard for some users, it broke it for any one using Victoria Metrics, and the current `GrafanaDashboard` manifest created by the Helm chart has no way to set the override needed to fix it.

The template should be updated to include a section to allow a user to set `.spec.datasources` with the following:
```
- datasourceName: VictoriaMetrics
  inputName: DS_PROMETHEUS
```

I've worked around it with a Kustomize patch in the meantime.

mengmann:
Fixes #1303

The `__inputs` format is only resolved by Grafana's UI import wizard, not by file/ConfigMap provisioning, causing all panels to fail with `Datasource ${DS_PROMETHEUS} was not found`.

Replaces `__inputs` with a proper `datasource` type variable in `templating.list`.

### DIFF
--- a/charts/spegel/monitoring/grafana-dashboard.json
+++ b/charts/spegel/monitoring/grafana-dashboard.json
@@ -1,14 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    }
-  ],
   "__elements": {},
   "__requires": [
     {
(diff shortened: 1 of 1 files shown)


phillebaba:
@mengmann thanks for the contribution :rose: This does fix the issue but I also want to make sure that we don't cause any regression in the future.

I think we should come up with a plan in #1303 first before making more changes. It would be great if we could together define what the requirements for a dashboard is so that it works in VictoriaMetrics, Grafana Cloud etc.

PR #1310, opened 2026-04-21
Reduce BaseTransport dial timeout

vflaux:
Drastically reduce the dial timeout for connections using `httpx.BaseTransport()`.
This is currently only used by `oic.NewClient()` to set up the HTTP client, and then by `Client.Fetch()` when fetching data from other Spegel pods.
The current timeout of 30s is far too long when a pod does not respond (for example, when it has been deleted).
As far as I understand, it’s possible to resolve the IP of a deleted pod via the P2P network, so this situation is expected to happen from time to time.
Correct me if I’m wrong, as I’m not familiar with the codebase and may have missed something.

### DIFF
--- a/pkg/httpx/httpx.go
+++ b/pkg/httpx/httpx.go
@@ -20,7 +20,7 @@ func BaseClient() *http.Client {
 func BaseTransport() *http.Transport {
 	return &http.Transport{
 		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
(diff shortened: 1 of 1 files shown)


phillebaba:
@vflaux could you please create a QA discussion or a specific bug report to discuss the problem you are facing at first? It will make reviewing any changes a lot easier.

I am closing this PR for now as it doesn't have any issue attached.

PR #1314, opened 2026-04-24
Add option to configure use of content events

brandond:
* *Alternative version of https://github.com/spegel-org/spegel/pull/1252*

Workaround for issues:
* https://github.com/spegel-org/spegel/issues/1096
* https://github.com/k3s-io/k3s/issues/13794

Adds a `WithContentEvents(bool)` option to the Containerd store that enables or disables use of ContentCreate events for advertising image content. Default continues to be true.

When disabled, content is advertised via ImageCreate events by walking the image manifest tree, instead of waiting for individual ContentCreate events

To be used in the following manner

```golang
storeOpts := []oci.ContainerdOption{
		oci.WithContentPath(filepath.Join(nodeConfig.Containerd.Root, "io.containerd.content.v1.content")),
		oci.WithContentEvents(false),
	}
```

cc @farazkhawaja

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -43,7 +43,8 @@ const (
 )
 
 type ContainerdConfig struct {
-	ContentPath string
+	ContentPath   string
+	ContentEvents bool
 }
 
 type ContainerdOption = option.Option[ContainerdConfig]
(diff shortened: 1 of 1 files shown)


farazkhawaja (approved):

phillebaba:
I don't think disabling one feature to enable another feature is the way to go. This makes content advertising slower to support advertising imported images. I would much rather see a complete solution that fixes the shortcomings on depending on content labels, which is used inconsistently by Containerd.

Additionally if we are going to support distributing imported images there needs to be integration tests added as Containerd treats images pulled and imported differently.

farazkhawaja:
@phillebaba should we eliminate the dependence on labels all together?

brandond:
> This makes content advertising slower to support advertising imported images.

Right but we're talking milliseconds here right? That seems like a small trade-off to make in exchange for actually being confident that all content is actually advertised, especially now that we KNOW that content labels are not guaranteed to be consistent with other stores.

The content event idea was nice, it may make for a nice reduction in overhead if containerd can eventually guarantee that they are consistent across content loading methods - but right now it doesn't seem to be usable for what Spegel actually needs to do.

PR #1317, opened 2026-04-27
Add percentile based timeout when racing fetches

### LINKED ISSUE #1311: Long dial timeout when fetching a blob to a remote peer
### Spegel version

0.7.0

### Kubernetes distribution

eks

### Kubernetes version

1.34

### CNI

cilium

### Describe the bug

While running Spegel in a cluster, I sometimes get dial timeout errors like the following:
```json
{
    "time": "xxx",
    "level": "ERROR",
    "source": {
        "function": "github.com/spegel-org/spegel/pkg/registry.(*Registry).raceFetch",
        "file": "spegel/pkg/registry/registry.go",
        "line": 478
    },
    "msg": "request to peer failed",
    "ref": "sha256:9824c27679d3b27c5e1cb00a73adb6f4f8d556994111c12db3c5d61a0c843df8",
    "path": "/v2/cilium/hubble-ui/blobs/sha256:9824c27679d3b27c5e1cb00a73adb6f4f8d556994111c12db3c5d61a0c843df8",
    "err": "after 30.037027758s, Get \"http://xxx:5000/v2/cilium/hubble-ui/blobs/sha256:9824c27679d3b27c5e1cb00a73adb6f4f8d556994111c12db3c5d61a0c843df8?ns=xxx\": dial tcp 100.88.168.139:5000: i/o timeout\nGet \"http://100.88.168.139:5000/v2/cilium/hubble-ui/blobs/sha256:9824c27679d3b27c5e1cb00a73adb6f4f8d556994111c12db3c5d61a0c843df8?ns=xxx\": dial tcp 100.88.168.139:5000: i/o timeout"
}
```
Note: this is a custom build that adds the time elapsed since the start of the request to the error.

This happens when a node does not have the image/blob requested via the registry (by containerd), while another node does. If the spegel pod on that other node is not responding (most likely because it was deleted), the dial attempt times out after 30 seconds before trying another pod. During this time, the containerd client waits.
Reducing this timeout would allow to try another peer sooner and reduce the overall image pull time in this scenario.

vflaux:
Add a race fetch timeout in `registry.go`.
Use twice the highest percentile duration from the hedger (with minimum of 100ms) and reset the timer each time a new fetch is started.
This does not affect content proxying, which still has no timeout.

The 100 ms minimum comes from test failures caused by very low hedger values during the tests, and setting a lower bound seems reasonable. This is still a fairly low value, and increasing it may be safer.
The timeout formula could be adjusted, or we could simply use a fixed value.

Fix #1311.

### DIFF
--- a/internal/resilient/hedge.go
+++ b/internal/resilient/hedge.go
@@ -26,6 +26,25 @@ func NewHedger(percentiles []float64, initial time.Duration) *Hedger {
 	}
 }
 
+// durationAtPercentile returns the duration at the given percentile, or the the initial duration if no data is available.
+func (h *Hedger) durationAtPercentile(percentile float64) time.Duration {
+	h.mx.RLock()
+	value := h.hist.ValueAtPercentile(percentile)
+	h.mx.RUnlock()
+	if value == 0 {
+		return h.initial
+	}
+	return time.Duration(value) * time.Millisecond
+}
+
+// HighestPercentileDuration returns the duration for the highest percentile.
+func (h *Hedger) HighestPercentileDuration() time.Duration {
+	if len(h.percentiles) == 0 {
+		return h.initial
+	}
+	return h.durationAtPercentile(h.percentiles[len(h.percentiles)-1])
+}
+
 // Size returns the amount of times a hedge channel will be triggered.
 func (h *Hedger) Size() int {
 	return len(h.percentiles)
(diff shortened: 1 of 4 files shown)


vflaux:
Added tests for `Hedger.HighestPercentileDuration()`.

phillebaba:
@vflaux is there a reason to not just reuse the idle timeout channel in the case where there are inflight fetches? Basically if there are no fetches we use the static resolve timeout but if there are fetches on going we use the double highest percentile?

phillebaba (approved):
Thanks for the hard work @vflaux 🌹

PR #1320, opened 2026-04-27
Fix data race when modifying range for failed blob request

### LINKED ISSUE #1318: Data race in happy eyeballs test
### Spegel version

v0.7.0

### Kubernetes distribution

N/A

### Kubernetes version

N/A

### CNI

N/A

### Describe the bug

Since the introduction of happy eyeballs the unit tests have become flaky.

https://github.com/spegel-org/spegel/actions/runs/24987265656/job/73163622881?pr=1305

This seems to be caused by some flaky data race detection that is occuring. I suspect that this isn't a bug but an actual data race that only happens in specific scenarios.

```
Read at 0x00c000332fa0 by goroutine 132:
  github.com/spegel-org/spegel/pkg/oci.(*Client).Fetch.func1()
      /home/runner/work/spegel/spegel/pkg/oci/client.go:315 +0x95e
  github.com/spegel-org/spegel/internal/resilient.Retry.func1()
      /home/runner/work/spegel/spegel/internal/resilient/retry.go:73 +0x3e
  github.com/spegel-org/spegel/internal/resilient.RetryValue[go.shape.interface {}]()
      /home/runner/work/spegel/spegel/internal/resilient/retry.go:99 +0x254
  github.com/spegel-org/spegel/internal/resilient.Retry()
      /home/runner/work/spegel/spegel/internal/resilient/retry.go:72 +0xe8
  github.com/spegel-org/spegel/pkg/oci.(*Client).Fetch()
      /home/runner/work/spegel/spegel/pkg/oci/client.go:302 +0xaf0
  github.com/spegel-org/spegel/pkg/registry.(*Registry).raceFetch.func2.1()
      /home/runner/work/spegel/spegel/pkg/registry/registry.go:430 +0x5c6
  github.com/spegel-org/spegel/pkg/httpx.HappyEyeballs[go.shape.struct { github.com/spegel-org/spegel/pkg/registry.rc io.ReadCloser; github.com/spegel-org/spegel/pkg/registry.desc github.com/opencontainers/image-spec/specs-go/v1.Descriptor; github.com/spegel-org/spegel/pkg/registry.peer github.com/spegel-org/spegel/pkg/routing.Peer }]()
      /home/runner/work/spegel/spegel/pkg/httpx/race.go:25 +0x187
  github.com/spegel-org/spegel/pkg/registry.(*Registry).raceFetch.func2()
      /home/runner/work/spegel/spegel/pkg/registry/registry.go:420 +0x30d

Previous write at 0x00c000332fa0 by goroutine 64:
  github.com/spegel-org/spegel/pkg/registry.(*Registry).mirrorHandler.func2()
      /home/runner/work/spegel/spegel/pkg/registry/registry.go:335 +0x1364
  github.com/spegel-org/spegel/pkg/registry.(*Registry).mirrorHandler()
      /home/runner/work/spegel/spegel/pkg/registry/registry.go:341 +0xb3e
  github.com/spegel-org/spegel/pkg/registry.(*Registry).registryHandler()
      /home/runner/work/spegel/spegel/pkg/registry/registry.go:219 +0xd73
  github.com/spegel-org/spegel/pkg/registry.(*Registry).registryHandler-fm()
      <autogenerated>:1 +0x48
  github.com/spegel-org/spegel/pkg/httpx.(*ServeMux).Handle.func1()
      /home/runner/work/spegel/spegel/pkg/httpx/mux.go:86 +0x566
  net/http.HandlerFunc.ServeHTTP()
      /opt/hostedtoolcache/go/1.26.2/x64/src/net/http/server.go:2286 +0x47
  github.com/spegel-org/spegel/pkg/httpx.(*ServeMux).ServeHTTP()
      /home/runner/work/spegel/spegel/pkg/httpx/mux.go:42 +0x421
  github.com/spegel-org/spegel/pkg/registry.TestRegistryHandler.func4()
      /home/runner/work/spegel/spegel/pkg/registry/registry_test.go:468 +0x4d0
  testing.tRunner()
      /opt/hostedtoolcache/go/1.26.2/x64/src/testing/testing.go:2036 +0x21c
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.26.2/x64/src/testing/testing.go:2101 +0x38
```

vflaux:
fix #1318 

Diagnostic & fix assisted by AI (GPT-5.3-Codex):
> Root Cause:
The race is on the mutable range object shared through DistributionPath.Range.
Read side: [client.go:315](https://github.com/spegel-org/spegel/blob/06c693120e23f372d6cbbc7e1cbc312ae21a8075/pkg/oci/client.go#L315) reads dist.Range.String() inside concurrent fetch/retry goroutines.
Write side: [registry.go:335](https://github.com/spegel-org/spegel/blob/06c693120e23f372d6cbbc7e1cbc312ae21a8075/pkg/registry/registry.go#L335) updates dist.Range.Start after a partial blob copy failure to resume from the next byte.
Why it races: DistributionPath is passed by value, but Range is a pointer field. So multiple goroutines end up sharing the same underlying Range object.

Solution is to pass a copy of the DistributionPath to fetch routines so they never race with the update at [registry.go:335](https://github.com/vflaux/spegel/blob/06c693120e23f372d6cbbc7e1cbc312ae21a8075/pkg/registry/registry.go#L335).
I don't think this bug affects responses by spegel - as it should only occurs in losing fetches - but this may still send requests with inconsistent ranges to other peers before being canceled.

### DIFF
--- a/pkg/httpx/range_test.go
+++ b/pkg/httpx/range_test.go
@@ -216,3 +216,66 @@ func TestContentRange(t *testing.T) {
 	_, err = ContentRangeFromRange(Range{Start: nil, End: nil}, 100)
 	require.EqualError(t, err, "start and end range cannot both be empty")
 }
+
+func TestRangeClone(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		rng  *Range
+		want *Range
+		name string
+	}{
+		{
+			name: "nil range",
+			rng:  nil,
+			want: nil,
+		},
+		{
+			name: "empty range",
+			rng:  &Range{},
+			want: &Range{},
+		},
+		{
+			name: "range with start and end",
+			rng: &Range{
+				Start: ptr.To(int64(0)),
+				End:   ptr.To(int64(100)),
+			},
+			want: &Range{
+				Start: ptr.To(int64(0)),
+				End:   ptr.To(int64(100)),
+			},
+		},
+		{
+			name: "range with only start",
+			rng: &Range{
+				Start: ptr.To(int64(50)),
+				End:   nil,
+			},
+			want: &Range{
+				Start: ptr.To(int64(50)),
+				End:   nil,
+			},
+		},
+		{
+			name: "range with only end",
+			rng: &Range{
+				Start: nil,
+				End:   ptr.To(int64(50)),
+			},
+			want: &Range{
+				Start: nil,
+				End:   ptr.To(int64(50)),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			rng := tt.rng.Clone()
+			if tt.rng != nil {
+				require.NotSame(t, tt.rng, rng)
+			}
+			require.Equal(t, tt.want, rng)
+		})
+	}
+}
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -266,18 +266,6 @@ func (r *Registry) mirrorHandler(ctx context.Context, dist oci.DistributionPath,
 		return
 	}
 
-	// Make copy to not modify distribution range.
-	if dist.Range != nil {
-		rng := &httpx.Range{}
-		if dist.Range.Start != nil {
-			rng.Start = ptr.To(*dist.Range.Start)
-		}
-		if dist.Range.End != nil {
-			rng.End = ptr.To(*dist.Range.End)
-		}
-		dist.Range = rng
-	}
-
 	// Retry requests until success or timeout.
 	for {
 		done := func() bool {
@@ -326,6 +314,7 @@ func (r *Registry) mirrorHandler(ctx context.Context, dist oci.DistributionPath,
 					log.Error(err, "copying of manifest data failed")
 					return true
 				case oci.DistributionKindBlob:
+					dist = dist.Clone()
 					if dist.Range == nil {
 						dist.Range = &httpx.Range{
 							Start: ptr.To(int64(0)),
--- a/pkg/httpx/range.go
+++ b/pkg/httpx/range.go
@@ -92,6 +92,26 @@ func (rng Range) String() string {
 	}
 }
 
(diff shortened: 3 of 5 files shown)


vflaux:
I added tests for new Clone() methods.

pkg/httpx/range_test.go
```diff
@@ -216,3 +215,66 @@ func TestContentRange(t *testing.T) {
 	_, err = ContentRangeFromRange(Range{Start: nil, End: nil}, 100)
 	require.EqualError(t, err, "start and end range cannot both be empty")
 }
+
+func TestRange_Clone(t *testing.T) {
```
phillebaba: Should be complete camelcase TestRangeClone

pkg/registry/registry.go
phillebaba, line 316: dist = dist.Clone()

pkg/registry/registry.go
```diff
@@ -414,20 +414,21 @@ func (r *Registry) raceFetch(ctx context.Context, iterator *routing.Iterator, di
 			fetchCtx, fetchCancel := context.WithCancel(ctx)
 			fetchCtxs[peer.Host] = fetchCtx
 			fetchCancels[peer.Host] = fetchCancel
+			distSnapshot := dist.Clone()
```
phillebaba: I think it is cleaner to do dist clone after line 328. Achieves the same thing but only clones the few times we actually have to. Then we can remove my bad range copy code while we are at it.

pkg/registry/registry.go
phillebaba: Remove 269 to 278

phillebaba (changes requested):
Thanks for taking the time to fix this! I like the clone solution a lot more as it is clearer. If you can implement my suggestion for where to clone we should be good to merge.

vflaux:
@phillebaba I've updated the PR to fix reports from the CI (I forgot to run tests and golanci-lint apparently).

phillebaba (approved):
LGTM 🌹

PR #1328, opened 2026-04-30
feat(mirror): support override_path on additionalMirrorTargets

### LINKED ISSUE #277: Document using Spegel with pull through registries
> After carefully reviewing the following containerd-related errors, I deleted the configurations related to "mirror" in `/etc/rancher/k3s/registries.yaml` and `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl`. Now, `spegel` is working properly.
> 
> ```shell
> time="2023-11-20T23:28:55.907708354+08:00" level=warning msg="failed to load plugin io.containerd.grpc.v1.cri" error="invalid plugin config: `mirrors` cannot be set when `config_path` is provided"
> ```

That's a bit unfortunate since the docs state:

> Spegel does not aim to replace projects like [Harbor](https://github.com/goharbor/harbor) or [Zot](https://github.com/project-zot/zot) but instead complements them.

I was hoping to use spegel in my cluster and zot as a pull thru cache deployed elsewhere. For what it's worth this is my current containerd mirrors:

```yaml
mirrors:
  docker.io:
    endpoint:
      - https://zot.domain.tld/v2/docker.io
  ghcr.io:
    endpoint:
      - https://zot.domain.tld/v2/ghcr.io
  quay.io:
    endpoint:
      - https://zot.domain.tld/v2/quay.io
  gcr.io:
    endpoint:
      - https://zot.domain.tld/v2/gcr.io
  registry.k8s.io:
    endpoint:
      - https://zot.domain.tld/v2/registry.k8s.io
  public.ecr.aws:
    endpoint:
      - https://zot.domain.tld/v2/public.ecr.aws
```

I don't see a way to have spegel take over this responsibility it seems like you either have spegel or a pull thru cache, maybe this can be a feature request?

_Originally posted by @onedr0p in https://github.com/XenitAB/spegel/issues/212#issuecomment-1830879118_

### LINKED ISSUE #708: support wildcard for repositories list

we have multi team working on one cluster, which team has there own private repository, which has different account, but same domain. for example: aws private repository - https://accountid.dkr.ecr.us-east-1.amazonaws.com. 

Is that possible to support wild card for repositories list? so we only need "dkr.ecr.us-east-1.amazonaws.com",  don't need input all accountid for this kind of scenarios.

_No response_

vanerac:
## Summary

Adds support for `override_path = true` on individual entries in `additionalMirrorTargets`. This makes Spegel usable as a transparent client in front of registries that expect a path prefix — most notably AWS ECR pull-through cache, where the upstream URL is shaped as `<account>.dkr.ecr.<region>.amazonaws.com/v2/<cache-prefix>`.

containerd's [hosts.toml](https://github.com/containerd/containerd/blob/main/docs/hosts.md#override_path-field) already supports `override_path = true` so that the configured URL is used literally, instead of containerd appending `/v2/<image>` to it. Spegel currently emits no such field, which means a target like `https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub` is silently turned into `https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub/v2/library/alpine/manifests/...` and the request fails.

This unblocks deployments that want to combine Spegel's peer-to-peer cache with ECR pull-through cache as a fallback target.

Closes #277.
Closes #708.

## Use case (ECR pull-through cache)

```yaml
spegel:
  mirroredRegistries:
    - https://docker.io
  additionalMirrorTargets:
    - url: https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub
      overridePath: true
```

## Before / after

`hosts.toml` rendered for the example above.

**Before** (existing behaviour — entry is rejected by `validateRegistryURL` because the path is non-empty):

```
invalid registry url path has to be empty: https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub
```

**After**:

```toml
server = 'https://docker.io'

[host.'http://127.0.0.1:5000']
capabilities = ['pull', 'resolve']
dial_timeout = '200ms'

[host.'https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub']
capabilities = ['pull', 'resolve']
dial_timeout = '200ms'
override_path = true
```

## Backwards compatibility

Plain string entries continue to work unchanged. The Helm chart accepts either:

```yaml
additionalMirrorTargets:
  - https://my-mirror.example.com         # plain string (existing)
  - url: https://my-ecr/v2/docker-hub     # new struct form
    overridePath: true
```

Internally:

- A new `MirrorTarget` type is introduced with a custom `UnmarshalJSON` that accepts both a string and an object form.
- `parseMirrorTargets` parses each `--mirror-targets` flag value as either a plain URL or a JSON object.
- `validateMirrorTargetURL` permits a non-empty path only when `overridePath` is set; the wildcard / mirrored-registry validator (`validateRegistryURL`) is unchanged.
- The Helm chart renders plain strings as-is, and renders struct-form entries as a JSON object string. The CLI flag remains a `[]string`, so existing users see no change.

## Test plan

- [x] `go test ./...` passes locally
- [x] `helm lint charts/spegel` passes
- [x] `helm template charts/spegel` correctly renders both forms
- [x] New unit tests cover both plain and override-path entries (`TestParseMirrorTargets`, `TestMirrorTargetUnmarshalJSON`, and a new `TestMirrorConfiguration` case asserting that `override_path = true` ends up in `hosts.toml`)
- [ ] Maintainers to confirm CI passes

### DIFF
--- a/charts/spegel/templates/daemonset.yaml
+++ b/charts/spegel/templates/daemonset.yaml
@@ -63,7 +63,11 @@ spec:
           - http://$(NODE_IP):{{ .Values.service.registry.nodePort }}
           {{- with .Values.spegel.additionalMirrorTargets }}
           {{- range . }}
+          {{- if kindIs "string" . }}
           - {{ . | quote }}
+          {{- else }}
+          - {{ toJson . | quote }}
+          {{- end }}
           {{- end }}
           {{- end }}
           - --resolve-tags={{ .Values.spegel.resolveTags }}
--- a/charts/spegel/values.yaml
+++ b/charts/spegel/values.yaml
@@ -162,8 +162,15 @@ spegel:
   mirroredRegistries: []
     # - https://docker.io
     # - https://ghcr.io
-  # -- Additional target mirror registries other than Spegel.
+  # -- Additional target mirror registries other than Spegel. Each entry is
+  # either a plain URL string, or an object with a url field and optional
+  # overridePath flag. overridePath maps to containerd's hosts.toml
+  # `override_path = true` and is required when the mirror URL contains a
+  # path prefix, for example AWS ECR pull-through cache.
   additionalMirrorTargets: []
+  #   - https://my-mirror.example.com
+  #   - url: https://123456789012.dkr.ecr.eu-west-1.amazonaws.com/v2/docker-hub
+  #     overridePath: true
   # -- Max amount of mirrors to attempt.
   mirrorResolveRetries: 3
   # -- Max duration spent finding a mirror.
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -443,7 +443,7 @@ func AddMirrorConfiguration(ctx context.Context, configPath string, mirroredRegi
 	if err != nil {
 		return err
 	}
-	parsedMirrorTargets, err := parseRegistries(mirrorTargets, false)
+	parsedMirrorTargets, err := parseMirrorTargets(mirrorTargets)
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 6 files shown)


phillebaba:
Thanks for the contribution @vanerac. Just wanted to confirm that you have read the [AI Policy](https://github.com/spegel-org/spegel/blob/main/AI_POLICY.md) before spending time reviewing this PR.

phillebaba:
Closing as issue should be discussed before implementation.

PR #1330, opened 2026-05-03
Migrate from stretchr/testify to go-openapi/testify/v2

### LINKED ISSUE #1316: Migrate to OpenAPI Testify
Testify is a great project striking a good balance between syntax sugar without adding too much complexity. It's v2 version has been under development for a long time, and would bring a lot of much needed changes. Especially when it comes to race conditions in eventually assertions.

OpenAPI has forked testify to develop what a v2 of testify would look like. It is already used in OpenAPI projects. It seems like the plan is to eventually upstream the changes, as the API is mostly compatible.
https://github.com/go-openapi/testify

I think that the OpenAPI version brings a lot of benefits that we want to access today, so it would be beneficial for Spegel to switch testing dependency and then evaluate if things change in the future.

simongottschlag:
Hey @phillebaba!

Full disclosure: AI was used to create these changes.

Swaps the test dependency from `github.com/stretchr/testify` v1.11.1 to `github.com/go-openapi/testify/v2` v2.5.0 across all three Go modules (root, `test/integration/containerd`, `test/integration/kubernetes`).

Performed with the upstream-recommended `migrate-testify` tool (pass 1, `--migrate`):

- Rewrote `assert` and `require` imports in 33 files.
- Renamed 10 `EventuallyWithT` call sites to `EventuallyWith`.
- Updated `go.mod` / `go.sum` in all three modules; ran `go mod tidy`.

No behavioural changes. `go build ./...`, `make lint`, and `make test-unit` (`-race`) pass. `go vet ./...` clean in both integration modules.

Out of scope (left for follow-up): generic-assertion upgrade (`migrate-testify --upgrade-generics`), goroutine-leak detection, `synctest`, colorized output, YAML enable.

Closes #1316


phillebaba (approved):
LGTM

PR #1374, opened 2026-06-01
Allow disabling internalTrafficPolicy for backwards compatibility

retributter:
## Details

As discussed [here](https://github.com/spegel-org/spegel/discussions/1371), creating a PR to add backwards compatibility, avoiding a double call for the [mirrorGo function](https://github.com/spegel-org/spegel/pull/18), when the layer isn't found because we didn't have time to decompress it from ctr.

This will add compatibility with `1.33.x` since preferSameNode was moved to `1.34` when `1.37` releases this will be removed from the chart as discussed previously

### DIFF
--- a/charts/spegel/templates/service.yaml
+++ b/charts/spegel/templates/service.yaml
@@ -33,9 +33,10 @@ metadata:
   {{- end }}
 spec:
   type: NodePort
-  trafficDistribution: PreferSameNode
   {{- if .Values.service.registry.internalTrafficLocal }}
   internalTrafficPolicy: Local
+  {{- else }}
+  trafficDistribution: PreferSameNode
   {{- end }}
   selector:
     app.kubernetes.io/component: spegel
--- a/charts/spegel/templates/service.yaml
+++ b/charts/spegel/templates/service.yaml
@@ -34,6 +34,9 @@ metadata:
 spec:
   type: NodePort
   trafficDistribution: PreferSameNode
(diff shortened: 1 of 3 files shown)


charts/spegel/templates/service.yaml
```diff
@@ -34,6 +34,9 @@ metadata:
 spec:
   type: NodePort
   trafficDistribution: PreferSameNode
```
phillebaba: I think this has to not be set for it to work?

phillebaba:
I remember there being some weird issue with specific CNI configurations and local node traffic when a Spegel instance is not running. Let me check this before merging.

charts/spegel/templates/service.yaml
retributter: You're right! 

Adapting it, I somehow thought this was already checked, but it's hardcoded and now depends on whether you're selecting the 'compatibility' for 1.33. 

When 1.37 is released, we can just come back to the old state. 

Letting the code here for future reference: 
https://github.com/spegel-org/spegel/commit/d2446b556dd8917c8521fe0ed7e86bad5e283401#diff-584cb0a65d152df3a041532d99b542f1ce9d2cb83d0e9b258277059f50e7bc16R36

retributter:
> I remember there being some weird issue with specific CNI configurations and local node traffic when a Spegel instance is not running. Let me check this before merging.

Waiting for your input then! :)

Issue #1378, opened 2026-06-09
HEAD on tag returns 200, GET on digest returns 404, and the image pull gets stuck in a retry loop


jvlerner:
### Spegel version

v0.7.1

### Kubernetes distribution

Kubeadm

### Kubernetes version

v1.32.10

### CNI

Calico

### Describe the bug

# Bug: HEAD on tag returns 200, GET on digest returns 404, and the image pull gets stuck in a retry loop

## Environment

- 32 nodes
- Kubernetes v1.32.10
- Ubuntu 22.04.5 LTS
- Kernel 5.15.0-161-generic
- containerd 1.7.29

## Summary

When an image is pulled on a node after Spegel has already started, Spegel advertises the tag in the DHT but not the manifest digest. Then, when another node tries to pull the same image, the flow looks like this:

1. `HEAD /manifests/:tag` returns `200 OK` because the tag is in the DHT and a peer is found.
2. `GET /manifests/:digest` returns `404` with `MANIFEST_UNKNOWN could not find peer for sha256:...`.
3. containerd stops the pull as soon as it gets that 404 while trying to fetch the manifest by digest. It does not fall back to the upstream registry.
4. kubelet then starts another pull attempt, so the same failing sequence keeps repeating and the pull ends up in a loop.

In practice, this breaks image pulls for runtime-pulled images in clusters using Spegel as a P2P mirror.

## Observed behavior

Containerd log on node `node-ltl9k`:

```text
time="2026-05-27T21:04:46.925Z" level=info msg="PullImage \"repository.intranet/...messaging:0.0.1-5f1a3d4\""
time="2026-05-27T21:04:47.910Z" level=error msg="PullImage failed" error="rpc error: code = NotFound desc = failed to pull and unpack image: failed to copy: httpReadSeeker: failed open: content at http://10...93:30020/v2/.../manifests/sha256:208eb7c...?ns=repository.intranet not found: not found"
```

Spegel log on node `node-ltl9k` (pod `spegel-6vh2m`):

```json
{"time":"2026-05-27T21:04:46.950Z","level":"INFO","path":"/v2/.../manifests/0.0.1-5f1a3d4","status":200,"method":"HEAD","latency":"22ms","handler":"mirror"}
[...]

PR #1379, opened 2026-06-09
Fix resolve digest pulls failing due to missing DHT advertisement and add fallback peer cache

jvlerner:
As part of the investigation for issue #1378, I built and tested a custom image with these changes (implemented using Claude Code Opus 4.6) in my environment.

The issue could no longer be reproduced after the fix, and image pulls completed successfully.


phillebaba:
Closing as issue was never discussed before contribution.

Issue #510, opened 2024-06-11
estargz-snapshotter support


phillebaba:
A question was raised by @ChrisBr in #176 if Spegel would support estargz-snapshotter. As it functions differently from SOCI it will require its own analysis to determine if it is possible to use with Spegel.

phillebaba:
Looks like the docs for mirror configuration are not up to date. However it seems like the snapshotter supports the newer CRI hosts mirror configuration. We can check the k3s PR for an example for how that would work.

https://github.com/containerd/stargz-snapshotter/issues/1647

At minimum we need to document this here as it is not documentated in stargz-snapshotter. I would like to add a test for this to verify that this is also working. The open question is whether we should add this check to verify to make sure that this configuration is set.

phillebaba:
Still need to test this but it seems like all that needs to be changed, assuming the snapshotter is properly configured, is to change the registry config slightly.

```toml
[plugins."io.containerd.grpc.v1.cri".registry]
  config_path = "/etc/containerd/certs.d"
```

Compared to.

```toml
[plugins."io.containerd.snapshotter.v1.stargz".registry]
  config_path = "/etc/containerd/certs.d"
```

phillebaba:
The latest stargz image uses Containerd v2 which Kind does not supporrt. We will need to wait until a full test  can be added.

https://github.com/kubernetes-sigs/kind/issues/3768

I did do some tests with an older  version of stargz and verifiedd that id does work with Spegel.

phillebaba:
After some more testing it is a bit more complicated than expected. Some things work while others do not. The main thing being that layers are stored differently. There is however an http cache which would be interesting to look at.

https://github.com/containerd/stargz-snapshotter/blob/dbddc6d7d0921ce66a9433726b546a3fe006e27b/fs/config/config.go#L40

simonepri:
@phillebaba any progress on this? Any idea what would be the right way to handle this?
I could find some time to contribute if we have a direction in mind already.

phillebaba:
I have not given any attention to this feature in the last year as I have limited time.

This is a pretty big feature that I would not use myself anywhere. There are a lot of small details that aren't obvious that make the feature harder than it first seems. Mostly around how snapshotters respect mirror configuration and store layers. I have raised some of these issues with the Containerd team, but that may also take time.

simonepri:
> I have raised some of these issues with the Containerd team, but that may also take time.

Is there some public trace of this?

[...]

PR #1386, opened 2026-06-12
Do not advertise content missing from the content store

simonepri:
Lazily pulled images have image records but no layer blobs in the content store, yet the [initial advertisement](https://github.com/spegel-org/spegel/blob/d1e7da21966cdf579913c8c3c6bf61b61900cb84/pkg/oci/containerd.go#L228-L257) announces every digest referenced by the manifest without checking that the blob exists. Peers resolving those layers get a 404.

The initial advertisement now only includes digests whose blobs are present in the content store. The check is only needed there: everything advertised after startup comes from [content create events](https://github.com/spegel-org/spegel/blob/d1e7da21966cdf579913c8c3c6bf61b61900cb84/pkg/oci/containerd.go#L299-L319), which fire when a blob is written and can never reference missing content.

The added integration test reproduces the lazy pull state by deleting a layer blob and fails without the fix.

Initial step toward supporting the stargz snapshotter, see #510.

### DIFF
--- a/pkg/oci/containerd.go
+++ b/pkg/oci/containerd.go
@@ -249,9 +249,13 @@ func (c *Containerd) Subscribe(ctx context.Context) (map[Image][]digest.Digest, 
 		}
 		contentIdx[cImg.Target.Digest] = refs
 
-		dgsts := []digest.Digest{}
-		for _, ref := range refs {
-			dgsts = append(dgsts, ref.Digest)
+		// The walk does not read leaf content like layers, so their existence has to be
+		// checked before advertising. Lazily pulled images have layers which are missing
+		// from the content store and cannot be served.
+		dgsts, err := c.existingDigests(ctx, refs)
+		if err != nil {
+			log.Error(err, "skipping image that cannot be checked for existing content", "image", img.String())
+			continue
 		}
 		initial[img] = dgsts
 	}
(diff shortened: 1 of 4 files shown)


brandond:
This sounds similar to the issue I was attempting to fix with https://github.com/spegel-org/spegel/pull/1314

phillebaba:
I am working on a long term solution for this that implements a content index. Fixes the challenges with imported layers while also enabling proper media type for blobs. Things are taking a bit longer as the time I have is sporadic. Once that is done we should have proper integration tests that verify imported content also works.

simonepri:
> we should have proper integration tests that verify imported content also works

I have new integration tests for this on my fork in case they are useful.

simonepri:
@phillebaba any update on this?

phillebaba:
I am doing some refactoring in #1464 that might fix this all together. It is a bit larger of a change which is why I haven't merged this yet.

simonepri:
> I have new integration tests for this on my fork in case they are useful.

In case it helps, they are here:
https://github.com/simonepri/spegel/commit/5b01251390e7911cc006b0cdab125d775ed04b9c

image

https://github.com/simonepri/spegel/commits/v0.18.3-estargz.1/

phillebaba:
@simonepri sorry I realized it was best to just merge this now and then rebase the changes to get the tests in. Is the PR in a good place now or do you want to do some adjustments?

simonepri:
I think it should be in a good state (modulo the conflict), I tested this on a live cluster at the time I wrote it (together with the other changes) and the integration seemed functional.

simonepri:
@phillebaba I've rebased the branch and checked, LGTM

farazkh-srti:
@phillebaba Hi, any plans on cutting a new spegel version?

PR #1387, opened 2026-06-12
Allow the registry namespace to be set with a header

simonepri:
The stargz snapshotter fetches layer chunks with ranged blob requests [built directly from its mirror configuration](https://github.com/containerd/stargz-snapshotter/blob/v0.18.2/fs/remote/resolver.go#L243-L248), without the `ns` query parameter that Containerd adds to mirror requests. Spegel [rejects any request without `ns`](https://github.com/spegel-org/spegel/blob/96e43bdd55d5ef6fc2ac72eeb284195633cc910a/pkg/oci/distribution.go#L109-L113), so today the snapshotter cannot use Spegel as a mirror at all.

The registry namespace can now also be set with the `OCI-Namespace` header, which Spegel already sets on mirror responses. The snapshotter sends the [headers from its mirror configuration](https://github.com/containerd/stargz-snapshotter/blob/v0.18.2/service/resolver/registry.go#L58-L59) on every blob request, and so does the snapshotter built into Containerd, which instead reads [the headers from hosts.toml](https://github.com/containerd/containerd/blob/v2.2.3/core/remotes/docker/config/hosts.go#L520-L522). The query parameter takes precedence over the header, leaving requests from Containerd unaffected.

A wildcard namespace is served without a registry, for clients configured with a catch all mirror which cannot name one. Digest references are self identifying so they are still served, while tag references keep requiring a registry. Requests which set neither the query parameter nor the header are rejected as before, and deployments which restrict the mirrored registries with `--mirrored-registries` keep only matching requests which have a registry.

Initial step toward supporting the stargz snapshotter, see #510.


phillebaba:
The `ns` parameter is part of the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#registry-proxying) so I would try to get this fixed upstream in estargz.

simonepri:
I will try but from a first look it's not a super trivial change on their side so it might take a while.

In the interim would it make sense to allow this via some experimental flag?

simonepri:
> I would try to get this fixed upstream in estargz.

Opened https://github.com/containerd/stargz-snapshotter/pull/2363 uptream

phillebaba:
@simonepri I am not sure how mirror config in estargz is managed, but I would be ok to check a header instead. This way the registry could be added to a header in the mirror config. The only downside is that you wont be able to use a wildcard registry configuration.

simonepri:
AFAIU using a header works, `MirrorConfig` has a [`header` field](https://github.com/containerd/stargz-snapshotter/blob/v0.18.2/service/resolver/registry.go#L58-L59) that gets sent on all the blob requests. Same for the builtin snapshotter in k3s, which reads mirrors from hosts.toml where headers are per host.

@phillebaba I've updated the PR to read `ns` first and fall back to `OCI-Namespace`, which we already set on responses, so a snapshotter mirror entry pointing at us would set a `OCI-Namespace = 'ghcr.io'` header.

The wildcard case is not really a problem either, for the catch all config the header would be set to `*`, which we already accept in `--mirrored-registries`.

Requests with neither `ns` nor the header keep being rejected either way.

simonepri:
@phillebaba https://github.com/containerd/stargz-snapshotter/pull/2363 got merged, do you think we should still merge this?

phillebaba:
I would prefer not adding another header as I would rather that clients implement the spec. Also it just adds complexity when we have no known client that would use this feature.

simonepri:
Ok closing then!

PR #1395, opened 2026-06-18
Refactor bootstrap certificate loading to enable unit tests

phillebaba:
Makes configuration a bit more simple with a single cert directory where files are loaded from.

This also removes the peer url in favor of a single bootstrap url. Avoiding some complexity.


phillebaba:
@dimitar-kostadinov just so you are aware of the changes I made to TLS parsing. Nothing that should have major impact on you but does remove some configuration complexity.

Issue #1411, opened 2026-06-30
Reduce ProviderAddrTTL to match provider record validity


patrickdomnick:
### Spegel version

v0.7.1

### Kubernetes distribution

EKS

### Kubernetes version

1.36.0

### CNI

AWS VPC CNI

### Describe the bug

ProviderAddrTTL is currently hardcoded at an hour.
When a peer is terminated, other nodes kept its IP address in their libp2p peerstore for up to 60 minutes. Whenever any of them is asked to look up a key the dead peer advertises the libp2p FindProviders flow still returns the dead peer's address from the peerstore cache.
That stale IP is then dialled, fails, and consumes retry budget on the mirror path.

### Community Rules

- [x] I agree to follow Spegels community rules and that all my communication is written by a human.

phillebaba:
Failed dials do not count towards the retry counter, it will just continue dialing other peers until one succeeds. The peer should not be returned after the record TTL expires. I dont see how reducing the address TTL would solve this problem?

The two things we could do is increase the peer limit in lookup and if possible rank peer addresses based on age.

Also please file bugs for the latest available release of Spegel. In this case it contains #1317 which fixes deadlock issues when racing for blobs.

phillebaba:
Closing as the retry budget is not affected during dial races.

PR #1412, opened 2026-06-30
fix: reduce ProviderAddrTTL to match provider record validity

patrickdomnick:
Reduce `ProviderAddrTTL` from a fixed 1 hour to `AdvertiseTTL + 2*MaxReprovideDelay + 5min` (24 minutes). This ensures dead peer addresses are evicted from the peerstore in a timeframe consistent with provider record expiry.
Stale-IP dials disappear within minutes of a peer going away instead of within an hour.

#1411


phillebaba:
Please read the contribution acceptance policy before submitting a PR immediately after creating an issue.

PR #1422, opened 2026-07-13
Reject range whose start is at or beyond content size

AruneshDwivedi:
A range like bytes=1000- against a 1000 byte blob passed validation and produced a content range with End set to one below the size, so Length() returned a negative Content-Length, while bytes=1500-2000 gave a zero length. RFC 9110 treats a start offset at or past the representation size as unsatisfiable, so ContentRangeFromRange now returns an error which the registry already maps to 416.

### DIFF
--- a/pkg/httpx/range.go
+++ b/pkg/httpx/range.go
@@ -142,6 +142,9 @@ func ContentRangeFromRange(rng Range, size int64) (ContentRange, error) {
 
 	// Offset with unknown size.
 	if rng.End == nil {
+		if *rng.Start >= size {
+			return ContentRange{}, fmt.Errorf("start %d is out of bounds for size %d", *rng.Start, size)
+		}
 		crng := ContentRange{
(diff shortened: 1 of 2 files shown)


pkg/httpx/range.go
```diff
@@ -142,6 +142,9 @@ func ContentRangeFromRange(rng Range, size int64) (ContentRange, error) {
 
 	// Offset with unknown size.
 	if rng.End == nil {
+		if *rng.Start >= size {
+			return ContentRange{}, fmt.Errorf("start %d is out of bounds for size %d", *rng.Start, size)
+		}
```
phillebaba: Remove the duplicate check and make it a single check before the end check.

phillebaba:
@AruneshDwivedi thanks for the quick fix, just a small NIT and it should be ready to merge.

phillebaba (approved):

Issue #562, opened 2024-08-16
Network policy example


jurim76:
### Describe the problem to be solved

Please provide network policy example or some requirement for network policy.
In a cluster with a default deny ingress policy, spegel doesn't work as expected, even with allow policy based on pod labels

phillebaba:
Different CNIs handle network policies applied on host networking differently. I do not have time to verify that this works with all the possible CNIs. Right now its up to the end user to deal with this. If you want to help documenting how to get network policies working with your specific setup I am happy to take contributions.

PR #1449, opened 2026-07-16
Add optional NetworkPolicy for the libp2p router to Helm chart

0xVijay:
This adds an optional NetworkPolicy to the chart that keeps inbound libp2p router traffic to Spegel peers within the release namespace.

The router (`--router-addr`, port 5001) runs the DHT and the peer mesh. Right now anything that can reach that port can join the mesh, which is more reach than it really needs, since the only things that legitimately talk to it are the other Spegel pods in the same namespace. This follows on from #562 where shipping a network policy example came up.

It is off by default (`networkPolicy.enabled: false`), so existing installs are untouched, and because policy behaviour still differs between CNIs. When enabled it leaves the registry and metrics ports open so image pulls and scraping keep working as before, and only constrains the router port. There is an `additionalRouterPeers` value for anyone who needs to let extra peers reach the router:

```yaml
networkPolicy:
  enabled: true
  additionalRouterPeers:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: monitoring
```

Ran `helm lint` and `helm template` with the policy on and off. Kubelet probes and other host-to-pod traffic follow the usual per-CNI behaviour.

### DIFF
--- a/charts/spegel/templates/networkpolicy.yaml
+++ b/charts/spegel/templates/networkpolicy.yaml
@@ -1,0 +1,37 @@
+{{- if .Values.networkPolicy.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "spegel.fullname" . }}
+  namespace: {{ include "spegel.namespace" . }}
+  labels:
+    {{- include "spegel.labels" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "spegel.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+  ingress:
+    # The libp2p router runs the DHT and peer mesh. Only other Spegel peers in
+    # this namespace need to reach it, so keep it closed to everything else.
+    - from:
+        - podSelector:
+            matchLabels:
+              {{- include "spegel.selectorLabels" . | nindent 14 }}
+        {{- with .Values.networkPolicy.additionalRouterPeers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      ports:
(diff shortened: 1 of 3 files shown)


charts/spegel/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,37 @@
...
+        {{- with .Values.networkPolicy.additionalRouterPeers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      ports:
+        - port: {{ .Values.service.router.port }}
+          protocol: TCP
+        - port: {{ .Values.service.router.port }}
+          protocol: UDP
+    # The registry and metrics ports stay open so image pulls and scraping keep
+    # working as before.
```
phillebaba: Remove comments from the policy/

charts/spegel/values.yaml
```diff
@@ -143,6 +143,12 @@ grafanaDashboard:
     matchLabels: {}
       # dashboards: grafana
 
+networkPolicy:
+  # -- If true creates a NetworkPolicy that limits libp2p router traffic to Spegel peers in the release namespace.
+  enabled: false
+  # -- Extra peers allowed to reach the libp2p router port, appended to the in-namespace Spegel selector.
+  additionalRouterPeers: []
```
phillebaba: Remove this as it just adds additional complexity. We don't have any use case where peers would join from outside the namespace.

phillebaba (changes requested):
Two samll NITs and then good to merge.

0xVijay:
Done, both make sense. Dropped the comments and removed `additionalRouterPeers` since there's no cross-namespace peer case.

phillebaba (approved):

PR #1501, opened 2026-09-01
Support images imported with ctr

### LINKED ISSUE #1096: Spegel does not advertise content loaded using `ctr image import`

### LINKED ISSUE #1267: Content advertisement depends on distribution.source labels that may not exist at event time
### Spegel version

v0.6.0

### Kubernetes distribution

k3s/other distros -- not distro specific issue

### Kubernetes version

latest versions

### CNI

-

### Describe the bug

Spegel's ContentCreate event handler calls contentLabelsToReferences() to resolve which registry/repo a blob belongs to. This reads containerd.io/distribution.source labels from the content store. These labels may not exist when the event fires.                                                    
                  
**Imported images (ctr image import / tarball)**                                                                                                                         
   
Labels are never set by containerd for imported content. The retry loop exhausts all 10 attempts and the event is dropped.
                                                                                                                                                
Consumers like k3s work around the missing digest by creating digest-bearing image references (e.g. name@sha256:...) after import. This triggers an ImageCreate event that takes the digest path, walks the full manifest tree, and discovers all layer digests — but the walk results are stored in contentIdx for delete tracking only  and nil events are returned. The layer information is discovered and then discarded.
                                                                                                                                                                       
**CRI-pulled images**
                                                                                                                                                                       
There is a race condition between content being committed (which fires ContentCreate) and distribution.source labels being applied. In some cases the labels are not present when the handler processes the event, causing individual blobs to never be advertised. This results in partial failures where most blobs in an image are advertised but some are silently dropped.

**Root cause**

contentLabelsToReferences() is the only mechanism to map digest → (registry, repository) at event time. It reads labels that containerd sets asynchronously and inconsistently across client implementations (CRI vs import vs nerdctl).

phillebaba:
This adds tests for importing images using ctr into the k8s.io namespace. Ensuring that the same Spegel functionality is available as it is for images pulled with crictl.

Closes #978
Closes #1252
Closes #1314 
Fixes #1267 
Fixes #1096

### DIFF
--- a/pkg/oci/client.go
+++ b/pkg/oci/client.go
@@ -32,7 +32,8 @@ const (
 )
 
 type ClientConfig struct {
-	TLSClientConfig *tls.Config
+	TLSClientConfig   *tls.Config
+	DisableKeepAlives bool
 }
 
 type ClientOption = option.Option[ClientConfig]
@@ -43,6 +44,13 @@ func WithTLS(rootCAs *x509.CertPool, certificates []tls.Certificate) ClientOptio
 			RootCAs:      rootCAs,
 			Certificates: certificates,
 		}
+		return nil
+	}
+}
+
+func WithDisableKeepAlives(v bool) ClientOption {
+	return func(cfg *ClientConfig) error {
+		cfg.DisableKeepAlives = v
 		return nil
 	}
 }
@@ -67,6 +75,7 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 	transport.MaxIdleConns = 100
 	transport.MaxConnsPerHost = 100
 	transport.MaxIdleConnsPerHost = 100
+	transport.DisableKeepAlives = cfg.DisableKeepAlives
 	httpClient.Transport = transport
 
 	ociClient := &Client{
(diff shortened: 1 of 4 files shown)


pkg/oci/containerd/containerd.go
```diff
@@ -347,6 +348,7 @@ func (c *Containerd) handleEvent(ctx context.Context, envelope events.Envelope,
 		}
 		return idx.AddImage(img), nil
 	case *eventtypes.ImageDelete:
+		time.Sleep(1 * time.Second)
```
phillebaba: This needs a better fix so wait for the diff.

PR #1512, opened 2026-09-04
Remove the Grafana dashboard

### LINKED ISSUE #1303: Grafana Dashboard broken for Victoria Metrics

phillebaba:
This removes the Grafana dashboard from the Helm chart. I have been asking for someone in the community to step-up and own the Grafana dashboard, without major success. Taking care of the dashboard takes time while also requiring a lot careful thought to match the needs of all end users. Instead it seems better for end users to create a dashboard to fit their needs, or to create a new community managed dashboard. Merging this would close #1303 as it isn't relevant anymore.


lemacbacon:
I think it's still a good idea to provide a dashboard directly from the chart but you're right it would more work to make it work on all the setups. 

I'll try to work on that when I'll have time if you're ok with that.

phillebaba:
The problem is that taking community contributions becomes a challenging task. What works for the contributor might break the dashboard for others. Has happened in the past. Once it breaks people show up asking to revert the changes previously made.

Dashboards exported from a Grafana instance tends to differ from instance to instance, meaning that every contribution makes a bunch of changes. There is no good way of enforcing any type of standard here. Something needs to change if we are going to keep the dashboard in the main repo.

PR #1524, opened 2026-09-21
Remove per-source IP connection limits from the libp2p resource manager

### LINKED ISSUE #1523: Pods cannot bootstrap on IPv6 clusters: libp2p resource manager treats the whole cluster as one /56
### Spegel version

0.7.4

### Kubernetes distribution

EKS

### Kubernetes version

1.36

### CNI

VPC 

### Describe the bug

I'm bringing up an IPv6-only cluster on EKS  (19 nodes, one VPC /56, all pods in 2600:xxxx:xxxx:xx00::/56) and it seems that libp2p's resource manager is exhibiting a gnarly rate limiting problem. The first few Spegel pods form a mesh, but every pod that starts after that fails its startup probe forever with `routing table is empty after bootstrapping`

**Diagnosis:** Go's libp2p's resource manager is rate limiting per-source IPv6 /56 to 8 concurrent connections and 0.2 new connections per second (burst 16), with only loopback exempted.  On IPv4 each pod is its own /32 bucket so this never shows. On IPv6 every pod in the cluster shares one /56, so each Spegel instance treats the entire cluster as a single source: bootstrap dials to the three lowest addresses get refused, the libp2p black hole detector then flips to Blocked, and the pod never joins 😭 

Repro: open 30 inbound connections from distinct addresses in one /56 against `rcmgr.NewResourceManager(rcmgr.NewFixedLimiter(rcmgr.DefaultLimits.AutoScale()))`; the ninth fails with `connections per ip limit exceeded`.

Proposed fix: disable this rate limiting in the resource manager. Spegel is used inside of a cluster and should not be doing per-subnet rate limiting!

AI Disclosure: I used Claude Fable 5.1 to track this down and fix it

### Community Rules

- [x] I agree to follow Spegels community rules and that all my communication is written by a human.

hyperb1iss:
Fixes #1523

go-libp2p's default resource manager protects public DHT nodes from a single source flooding them: every IPv6 /56 is treated as one source and allowed 8 concurrent connections and 0.2 new connections per second, with only loopback exempt. Spegel only peers with pods in its own cluster, and clusters with IPv6 pod networking place every pod in a single /56, so each Spegel instance treats the whole cluster as one source. Bootstrap dials are refused with "rate limit exceeded" once a handful of pods are up, and any pod that starts later never joins the mesh 😭 

Build the resource manager with the default scaled limits but without the per subnet connection caps and the connection rate limiter. The router test already replaced the resource manager to get past the same limit, so it now runs on the production one.

AI use: I used Fable 5.1 to find and fix this issue. It definitely takes away the fun and satisfaction one gets after a successful week-long deep dive bug hunt into the bowels of EKS, CNI, and Go networking libraries. Fable took about 20m to figure it out 🤷🏽‍♀️


phillebaba (approved):
Thanks for looking into this. I new about the limits from when I wrote the unit tests but I wasn't aware it was affecting IPv6 worse than IPv4.

